PR #3, opened 2015-10-06
Adds support for system_clock time_points of any duration.

devjgm:
This change removes the need for a 128-bit count of nanoseconds that was
defined by the cctz::time_point type. Instead, cctz functions that
accept time_points can accept any time_point of the form:

  `std::chrono::time_point<std::chrono::system_clock, D>`

And functions that produce time points, such as MakeTime() and
`MakeTimeInfo()` now return system_clock time points that are a 64-bit
count of seconds.

There is still a "cctz::time_point" alias defined, but that is simply a
shorthand, and it is not defining a new concept.

The advantages of this change are:
- No unnecessary "cctz::time_point" concept.
- No required use of int128 in the interface.
- All time points that users already have are now accepted.
- No need to pass 128-bits of data where the low 64-bits are all zeros.

Misc other changes:
- Added warning comment about Breakdown's offset, is_dst, and abbr
  fields.
- Removed the subseconds field from Breakdown. We discovered that it is
  rarely needed (and when used, it's often used incorrectly). We can add
  it back later if/when needed.

### DIFF
--- a/src/cctz.h
+++ b/src/cctz.h
@@ -37,6 +37,7 @@
 #include <chrono>
 #include <cstdint>
 #include <string>
+#include <utility>
 
 namespace cctz {
 
@@ -297,32 +298,29 @@ bool Parse(const std::string& format, const std::string& input,
 namespace cctz {
 
 namespace internal {
-// Floors tp to a second boundary. Optionally returns subseconds.
-template <typename D, typename Subseconds>
-inline time_point<seconds64> FloorSeconds(const time_point<D>& tp,
-                                          Subseconds* subseconds) {
+// Floors tp to a second boundary and sets *subseconds.
+template <typename D>
+inline std::pair<time_point<seconds64>, D>
+FloorSeconds(const time_point<D>& tp) {
   auto sec = std::chrono::time_point_cast<seconds64>(tp);
   auto sub = tp - sec;
   if (sub.count() < 0) {
     sec -= seconds64(1);
     sub += seconds64(1);
   }
-  if (subseconds) *subseconds = std::chrono::duration_cast<Subseconds>(sub);
-  return sec;
+  return {sec, std::chrono::duration_cast<D>(sub)};
 }
-// Specialization for when tp is already second aligned.
-template <typename Subseconds>
-inline time_point<seconds64> FloorSeconds(const time_point<seconds64>& sec,
-                                          Subseconds* subseconds) {
-  if (subseconds) *subseconds = Subseconds::zero();
-  return sec;
+// Overload for when tp is already second aligned.
+inline std::pair<time_point<seconds64>, seconds64>
+FloorSeconds(const time_point<seconds64>& tp) {
+  return {tp, seconds64(0)};
 }
 }  // namespace internal
 
 template <typename D>
 inline Breakdown BreakTime(const time_point<D>& tp, const TimeZone& tz) {
   Breakdown BreakTime(const time_point<seconds64>&, const TimeZone&);
-  return BreakTime(internal::FloorSeconds(tp, static_cast<D*>(nullptr)), tz);
+  return BreakTime(internal::FloorSeconds(tp).first, tz);
 }
 
 template <typename D>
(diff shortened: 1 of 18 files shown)


devjgm:
@devbww PTAL at this diff and pull request. It's the implementation of what we discussed yesterday. Note that I did end up removing subseconds from BreakTime(). After a refactoring, I was able to get that in the one place we needed (Format()) via a separate helper function, so I removed that out-param from BreakTime. As we learned yesterday, this field doesn't seem to be very necessary anyway. We can always consider adding it back IFF we see that it's necessary and we have a nice place to put it.

src/cctz.h
```diff
@@ -91,35 +96,51 @@ TimeZone LocalTimeZone();
...
+  int month;         // month of year [1:12]
+  int day;           // day of month [1:31]
+  int hour;          // hour of day [0:23]
+  int minute;        // minute of hour [0:59]
+  int second;        // second of minute [0:59]
+  int weekday;       // 1==Mon, ..., 7=Sun
+  int yearday;       // day of year [1:366]
+
+  // Note: The following fields exist for backward compatibility with older
+  // APIs. Accessing these fields directly is a sign of misguided logic in the
```
devbww: "misguided" sounds a little strong.  How about "imprudent"?

devjgm: Much better. Changed to imprudent.

src/cctz.h
```diff
@@ -249,8 +278,77 @@ std::string Format(const std::string& format, const time_point& tp,
...
+}  // namespace cctz
+
+//
+// IMPLEMENTATION DETAILS
+//
+namespace cctz {
+
+namespace internal {
+// Floors tp to a second boundary. Optionally returns subseconds.
+template <typename D, typename Subseconds = std::chrono::nanoseconds>
```
devbww: Why nanoseconds?  It's arbitrary and only applies in the nullptr case, right?  D::duration perhaps?

src/cctz.h
```diff
@@ -249,8 +278,77 @@ std::string Format(const std::string& format, const time_point& tp,
...
+  Breakdown BreakTime(const time_point<seconds64>&, const TimeZone&);
+  return BreakTime(internal::FloorSeconds(tp), tz);
+}
+
+template <typename D>
+inline std::string Format(const std::string& format, const time_point<D>& tp,
+                          const TimeZone& tz) {
+  std::string Format(const std::string&, const time_point<seconds64>&,
+                     const std::chrono::nanoseconds&, const TimeZone&);
+  std::chrono::nanoseconds ns{};
```
devbww: Remember that we need to generalize (or at least expand) nanoseconds here and below.

src/cctz.h
```diff
@@ -249,8 +278,77 @@ std::string Format(const std::string& format, const time_point& tp,
...
+//
+// IMPLEMENTATION DETAILS
+//
+namespace cctz {
+
+namespace internal {
+// Floors tp to a second boundary. Optionally returns subseconds.
+template <typename D, typename Subseconds = std::chrono::nanoseconds>
+inline time_point<seconds64> FloorSeconds(const time_point<D>& tp,
+                                          Subseconds* subseconds = nullptr) {
```
devbww: Is defaulting the second parameter worth it?

src/cctz.h
devjgm: Yes, it was something arbitrary to allow `FloorSeconds()` to be called with a single parameter when the subseconds were not needed. I changed it to default to D.

src/cctz.h
devjgm: Correct. This is all internal to the implementation and it preserves existing behavior. In a follow up change we can expand the resolution of `Format()` and `Parse()`. Changing that now seemed unnecessary and I'd rather do separate, smaller changes.

src/cctz_fmt.cc
```diff
@@ -657,8 +654,9 @@ bool Parse(const std::string& format, const std::string& input,
   if (*data != '\0') return false;
 
   // If we saw %s then we ignore anything else and return that time.
-  if (saw_precent_s) {
-    *tpp = time_point(std::chrono::duration<int64_t>(percent_s_time));
+  if (saw_percent_s) {
+    *tpp = FromUnixSeconds(percent_s_time);
+    *ns = {};
```
devbww: I can't see any guarantee that this gives the same value as zero().

src/cctz.h
devjgm: The only difference would be that the body of `BreakTime()` below would become:

```diff
-  return BreakTime(internal::FloorSeconds(tp), tz);
+  return BreakTime(internal::FloorSeconds(tp, static_cast<D*>(nullptr)), tz);
```
Since defaulting the second parameter doesn't cost anything, it seemed reasonable. It doesn't matter either way to me. I'll change it if you prefer.

src/cctz_fmt.cc
devjgm: OK. Changed to use zero() explicitly.

devbww:
LGTM

PR #12, opened 2015-11-12
Add googletest -inl headers to BUILD

bsilver8192:
With Bazel's new C++ sandboxing, it doesn't build without this.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -4,6 +4,7 @@ cc_library(
     srcs = glob(
         [
             "google*/src/*.cc",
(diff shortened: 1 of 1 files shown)


bsilver8192:
#13 is a nicer fix.

devjgm:
JFYI, I'm not ignoring this, I've just been sick. I'll be back soon. Thanks.

bsilver8192:
Thanks, but you actually should ignore this one. :) Merge #13 instead.

PR #13, opened 2015-11-13
Add gtest-internal-inl.h to textual_headers of gtest.

lberki:
This makes it possible to compile cctz with the Bazel sandbox.


davidzchen:
Yes, Bazel now requires that all header files be declared. Here is the discussion for this particular issue: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/bazel-discuss/I7YXhCrmwx0/QmVgyTQSAwAJ

PR #15, opened 2015-12-11
MS Visual Studio support is realized.

SemenMartynov:
It is possible to use this library with tzdata on MS Windows. Path to the tzdata should be set via environment variables.

### DIFF
--- a/src/cctz.h
+++ b/src/cctz.h
@@ -317,27 +317,27 @@ FloorSeconds(const time_point<seconds64>& tp) {
 }
 }  // namespace internal
 
+Breakdown BreakTime(const time_point<seconds64>&, const TimeZone&);
 template <typename D>
 inline Breakdown BreakTime(const time_point<D>& tp, const TimeZone& tz) {
-  Breakdown BreakTime(const time_point<seconds64>&, const TimeZone&);
   return BreakTime(internal::FloorSeconds(tp).first, tz);
 }
 
+std::string Format(const std::string&, const time_point<seconds64>&,
+	const std::chrono::nanoseconds&, const TimeZone&);
 template <typename D>
 inline std::string Format(const std::string& format, const time_point<D>& tp,
                           const TimeZone& tz) {
-  std::string Format(const std::string&, const time_point<seconds64>&,
-                     const std::chrono::nanoseconds&, const TimeZone&);
   const auto p = internal::FloorSeconds(tp);
   const auto n = std::chrono::duration_cast<std::chrono::nanoseconds>(p.second);
   return Format(format, p.first, n, tz);
 }
 
+bool Parse(const std::string&, const std::string&, const TimeZone&,
+	time_point<seconds64>*, std::chrono::nanoseconds*);
 template <typename D>
 inline bool Parse(const std::string& format, const std::string& input,
                   const TimeZone& tz, time_point<D>* tpp) {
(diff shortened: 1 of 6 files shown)


src/cctz.h
```diff
@@ -317,27 +317,27 @@ FloorSeconds(const time_point<seconds64>& tp) {
 }
 }  // namespace internal
 
+Breakdown BreakTime(const time_point<seconds64>&, const TimeZone&);
```
SemenMartynov: I do not like that these functions become available globally, but otherwise I get ugly linker errors. Maybe someone will tell me better solution?

src/cctz_info.cc
```diff
@@ -626,9 +630,22 @@ bool TimeZoneInfo::Load(const std::string& name) {
...
+  while ((pos = path.find("/", pos)) != std::string::npos) {
+    path.replace(pos, 1, "\\");
+    pos += 2;
+  }
+#endif // _WIN32 || _WIN64
+
   // Load the time-zone data.
   bool loaded = false;
+#ifdef _WIN32 || _WIN64
+  if (FILE* fp = fopen(path.c_str(), "rb")) {
```
mwinterb: Can't "rb" be used as the open mode regardless of platform to avoid cluttering this part with the #ifdef?

mwinterb:
All of the #ifdef's need to be changed to something similar to #if defined(_WIN32) || defined(_WIN64) the " || _WIN64" is ignored by the preprocessor.

SemenMartynov:
Updated according to the comments.

src/cctz_libc.cc
```diff
@@ -20,6 +20,7 @@
 #include <ctime>
 
 namespace cctz {
+#if defined(_WIN32) || defined(_WIN64)
```
mwinterb: Sorry, I missed that these were #ifndef's, so the ones in the cctz_libc files will have to be #if !(defined(_WIN32) || defined(_WIN64)).

But what's actually preventing these from working in Visual Studio? Is it just tm_gmtoff and tm_zone? I think tm_gmtoff is indirectly accessible through _isindst, _getdstbias, and _gettimezone. tm_zone is returned by _get_tzname, but that doesn't return the time zone abbreviation. It is what VS's strftime prints for %Z so cctz::Format would be consistent with the "native" strftime, if that's acceptable. Windows doesn't seem to have the ability to retrieve the abbreviation for a specific time zone.

SemenMartynov: >>Sorry, I missed that these were #ifndef's, so the ones in the cctz_libc files will have to be #if !(defined(_WIN32) || defined(_WIN64)).

Sorry, it's my fault... It's almost morning in my time zone)

devbww:
Thanks Semen.  We appreciate having someone trying this out on Windows.

I believe we have addressed the issues, using variants of the code you submitted.  Hopefully you can try to build again and let us know what we messed up.

We did leave out one thing though---the "Switching Linus-path to Windows-path" translation of / to \.  Are you sure this is necessary?  Our understanding is that something in fopen() or below should already be taking care of this.

Thanks again.

devbww:
This pull request is now obsolete as we've moved to v2.0.  Hopefully the code now compiles cleanly under Windows, but we're certainly interested in any fixes/updates for the Windows environment.

PR #17, opened 2016-02-01
Feature msvs support

anastasia0121:
It is necessary to solve the problems with export and import functions and transmission paths of tzdata to compile on Windows.


devbww:
This pull request is now obsolete as we've moved to v2.0.  We have not yet addressed the "export" issue however, but are happy to explore it with you in the new code base.  Thanks.

PR #30, opened 2016-10-25
Makefile Shared Library

mattyclarkson:
This adds shared library support in the makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ $(CCTZ_SHARED_LINK): $(CCTZ_SHARED_SONAME)
 $(CCTZ_LIB): $(CCTZ_OBJS)
 	$(AR) $(ARFLAGS) $@ $(CCTZ_OBJS)
 
-install: install_hdrs install_lib
+install: install_hdrs install_lib install_shared_lib
 
 install_hdrs: $(CCTZ_HDRS)
 	$(SUDO) $(MKDIR) -p $(DESTDIR)$(PREFIX)/include
@@ -97,6 +97,10 @@ install_hdrs: $(CCTZ_HDRS)
 install_lib: $(CCTZ_LIB)
 	$(SUDO) $(MKDIR) -p $(DESTDIR)$(PREFIX)/lib
 	$(SUDO) $(CP) -p $? $(DESTDIR)$(PREFIX)/lib
+
+install_shared_lib: $(CCTZ_SHARED_LIB) $(CCTZ_SHARED_SONAME) $(CCTZ_SHARED_LINK)
+	$(SUDO) $(MKDIR) -p $(DESTDIR)$(PREFIX)/lib
+	$(SUDO) $(CP) -pd $? $(DESTDIR)$(PREFIX)/lib
 
 clean:
 	@$(RM) -r $(EXAMPLES:=.dSYM) $(EXAMPLES:=.o) $(EXAMPLES:=.d) $(EXAMPLES)
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ CXX = g++
 STD = c++11
 OPT = -O
 PREFIX = /usr/local
+DESTDIR =
 
 # possible support for googletest
 ## TESTS = civil_time_test time_zone_lookup_test time_zone_format_test
@@ -74,12 +75,12 @@ $(CCTZ_LIB): $(CCTZ_OBJS)
 install: install_hdrs install_lib
 
 install_hdrs: $(CCTZ_HDRS)
-	$(SUDO) $(MKDIR) -p $(PREFIX)/include
-	$(SUDO) $(CP) -p $? $(PREFIX)/include
+	$(SUDO) $(MKDIR) -p $(DESTDIR)$(PREFIX)/include
+	$(SUDO) $(CP) -p $? $(DESTDIR)$(PREFIX)/include
 
 install_lib: $(CCTZ_LIB)
-	$(SUDO) $(MKDIR) -p $(PREFIX)/lib
-	$(SUDO) $(CP) -p $? $(PREFIX)/lib
+	$(SUDO) $(MKDIR) -p $(DESTDIR)$(PREFIX)/lib
+	$(SUDO) $(CP) -p $? $(DESTDIR)$(PREFIX)/lib
 
 clean:
 	@$(RM) -r $(EXAMPLES:=.dSYM) $(EXAMPLES:=.o) $(EXAMPLES:=.d) $(EXAMPLES)
--- a/Makefile
+++ b/Makefile
@@ -39,12 +39,19 @@ ARFLAGS = rcs
 LDFLAGS = -pthread
 LDLIBS = $(TEST_LIBS)
 
+LN = ln
 CP = cp
 MKDIR = mkdir
 SUDO =
 
+CCTZ_VERSION_MAJOR = 2
+CCTZ_VERSION_MINOR = 0
+
 CCTZ = cctz
 CCTZ_LIB = lib$(CCTZ).a
+CCTZ_SHARED_LINK = lib$(CCTZ).so
+CCTZ_SHARED_SONAME = $(CCTZ_SHARED_LINK).$(CCTZ_VERSION_MAJOR)
+CCTZ_SHARED_LIB = $(CCTZ_SHARED_SONAME).$(CCTZ_VERSION_MINOR)
 
 CCTZ_HDRS =			\
 	civil_time.h		\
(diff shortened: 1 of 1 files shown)


pau-climent:
I've just applied for/signed a CLA. How can this now be solved?

Makefile
```diff
@@ -69,12 +70,12 @@ $(CCTZ_LIB): $(CCTZ_OBJS)
 install: install_hdrs install_lib
 
 install_hdrs: $(CCTZ_HDRS)
-	$(SUDO) mkdir -p $(PREFIX)/include
-	$(SUDO) cp -p $? $(PREFIX)/include
+	$(SUDO) mkdir -p $(DESTDIR)/$(PREFIX)/include
```
devbww: What is it that DESTDIR is buying us over PREFIX?  They always appear together.  Should we just rename PREFIX to DESTDIR?

Makefile
```diff
@@ -38,10 +38,17 @@ CPPFLAGS = -Wall -I$(SRC)include -std=$(STD) -pthread \
...
+LN = ln
+CP = cp
+MKDIR = mkdir
 
 SUDO =
 
 CCTZ_LIB = libcctz.a
+CCTZ_STATIC_LIB = $(CCTZ_LIB)
+CCTZ_SHARED_LIB = libcctz.so.2.0
+CCTZ_SHARED_SONAME = $(basename $(CCTZ_SHARED_LIB))
```
devbww: Why use basename on paths that do not contain slashes?

Makefile
```diff
@@ -62,25 +69,35 @@ EXAMPLES = classic epoch_shift hello example1 example2 example3 example4
 
 all: $(TESTS) $(TOOLS) $(EXAMPLES)
 
-$(TESTS) $(TOOLS) $(EXAMPLES): $(CCTZ_LIB)
+$(TESTS) $(TOOLS) $(EXAMPLES): $(CCTZ_SHARED_LIB)
 
-$(CCTZ_LIB): $(CCTZ_OBJS)
+$(CCTZ_SHARED_LIB): $(CCTZ_OBJS) $(CCTZ_SHARED_SONAME)
```
devbww: There seems to be some circularity here ... CCTZ_SHARED_LIB depends on CCTZ_SHARED_SONAME and CCTZ_SHARED_SONAME depends on CCTZ_SHARED_LIB.

Makefile
```diff
@@ -38,10 +38,17 @@ CPPFLAGS = -Wall -I$(SRC)include -std=$(STD) -pthread \
...
 LDLIBS = $(TEST_LIBS)
+LN = ln
+CP = cp
+MKDIR = mkdir
 
 SUDO =
 
 CCTZ_LIB = libcctz.a
+CCTZ_STATIC_LIB = $(CCTZ_LIB)
+CCTZ_SHARED_LIB = libcctz.so.2.0
```
devbww: Is "2.0" referring to "CCTZ Version 2"?  If so, perhaps we should be providing some more central versioning information.

Makefile
```diff
@@ -42,6 +42,7 @@ LDLIBS = $(TEST_LIBS)
 SUDO =
 
 CCTZ_LIB = libcctz.a
+CCTZ_STATIC_LIB = $(CCTZ_LIB)
```
mattyclarkson: Yeah, makes space for `CCTZ_SHARED_LIB`. The `CCTZ_LIB` was just left there as backwards compatible variable in case anyone was doing `make CCTZ_LIB=libmycctz.a` for some reason. Happy to remove it!

Makefile
mattyclarkson: It removes the last part of the version number, so it becomes `libcctz.so.2` (drops the `.0`)

Makefile
mattyclarkson: Sure, usually a `VERSION` file is used. Would happily read that file to determine the version. I initially did a describe of the lastest `git tag` but realised that it wouldn't work for archives.

Makefile
mattyclarkson: `DESTDIR` is a common override that is used in `autotools` style makefiles. See the [documentation](https://www.gnu.org/software/automake/manual/html_node/DESTDIR.html).

The reason to have a seperate `PREFIX` and `DESTDIR` is that the prefix is the installation prefix but the DESTDIR allows installation into a different location. This is especially useful when cross-compiling as the `PREFIX` is actually `/usr` but the `DESTDIR` is something like `/home/user/cc-staging`. Then the `DESTDIR` is packaged up and expanded onto the target device.

Some projects use the `prefix` variable to understand where the project is to be installed eventually. This project doesn't use that at all in the C++ code but it would be nice if we can play nice with common `Makefile` conventions. Happy to remove it if you don't want it though :smile:

Makefile
devbww: Let's just leave CCTZ_LIB as it is then.  That is, forget about CCTZ_STATIC_LIB ... just introduce CCTZ_SHARED_LIB.

Makefile
devbww: OK, thanks.  Let's just leave it as a literal "2.0" for now, then.

Makefile
devbww: OK, then let's at least remove the /'s between $(DESTDIR) and $(PREFIX).

Makefile
mattyclarkson: Whoops, sorry that's not a dependency.

mattyclarkson:
@devbww should be good to go unless you have extra comments.

mattyclarkson:
Once this goes in; are there any plans to do another release? It would be really helpful to have an archive that we can download and integrate into our distribution that can build shared libraries directly from the makefile. We currently are running with patches on top of the v2.0 tarball.

Makefile
```diff
@@ -69,6 +73,15 @@ $(TESTS) $(TOOLS) $(EXAMPLES): $(CCTZ_LIB)
 
 $(TESTS:=.o) $(TOOLS:=.o) $(EXAMPLES:=.o):
 
+$(CCTZ_SHARED_LIB): $(CCTZ_OBJS)
+	$(CC) $(LDFLAGS) -shared -Wl,-soname,$(CCTZ_SHARED_SONAME) -o$@ $(CCTZ_OBJS)
+
+$(CCTZ_SHARED_SONAME): $(CCTZ_SHARED_LIB)
```
devbww: Is it important for these links to be in the build tree, or are they best handled by the installation rule?

Makefile
```diff
@@ -67,7 +67,7 @@ CCTZ_OBJS =			\
 TOOLS = time_tool
 EXAMPLES = classic epoch_shift hello example1 example2 example3 example4
 
-all: $(TESTS) $(TOOLS) $(EXAMPLES)
+all: $(TESTS) $(TOOLS) $(EXAMPLES) $(CCTZ_SHARED_LINK)
```
devbww: Why is this required?

Makefile
```diff
@@ -39,12 +39,16 @@ ARFLAGS = rcs
...
 LDLIBS = $(TEST_LIBS)
 
+LN = ln
 CP = cp
 MKDIR = mkdir
 SUDO =
 
 CCTZ = cctz
 CCTZ_LIB = lib$(CCTZ).a
+CCTZ_SHARED_LIB = lib$(CCTZ).so.2.0
```
devbww: Perhaps the GNU dependency is already there, with VPATH and -include, but perhaps it is still clearer this way.

devbww:
> Once this goes in; are there any plans to do another release? It would be really helpful to have an archive that we can download and integrate into our distribution that can build shared libraries directly from the makefile. We currently are running with patches on top of the v2.0 tarball.

Yes, we can do that after a few other pending changes land.  Thanks.

Makefile
mattyclarkson: It's not. Was just so that the shared library ended up in the build folder when doing `make all`. Will remove it.

Makefile
mattyclarkson: Nope. They can be done by the install. Will do that.

Makefile
mattyclarkson: I added a `CCTZ_VERSION_MAJOR` and `CCTZ_VERSION_MINOR` variable to make it a bit more obvious what those numbers are. Also allows those two variables to be overridden.

Makefile
mattyclarkson: Hmm, having these rules makes it a bit easier to build the library as you can do `make libcctz.so` rather than having to know the version number (`make libcctz.so.2.0`). I've left them in but I can remove them if you don't see that as a benefit.

Makefile
```diff
@@ -69,6 +73,15 @@ $(TESTS) $(TOOLS) $(EXAMPLES): $(CCTZ_LIB)
 
 $(TESTS:=.o) $(TOOLS:=.o) $(EXAMPLES:=.o):
 
+$(CCTZ_SHARED_LIB): $(CCTZ_OBJS)
+	$(CC) $(LDFLAGS) -shared -Wl,-soname,$(CCTZ_SHARED_SONAME) -o$@ $(CCTZ_OBJS)
```
mattyclarkson: Done.

PR #44, opened 2017-03-28
Fix warning: comparison between signed and unsigned integer expressions

proller:
gcc 7.0.1:
```
libcctz/src/time_zone_info.cc: In member function 'bool cctz::TimeZoneInfo::Load(const string&, FILE*)':
libcctz/src/time_zone_info.cc:447:18: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     while (index != hdr.typecnt && transition_types_[index].is_dst)
            ~~~~~~^~~~~~~~~~~~~~
libcctz/src/time_zone_info.cc:449:15: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     if (index != hdr.typecnt)
         ~~~~~~^~~~~~~~~~~~~~
```


devbww:
It appears to me that making the proposed change will elicit the same error from other compilers, and that any fix will need to be more subtle.

devbww:
I was also unable to reproduce the -Wsign-compare error using x86-64 gcc 7 with a small example (where std::size_t is unsigned long and std::uint_fast8_t is unsigned char) on gcc.goldbolt.org.  Changing to std::int_fast8_t (as mentioned above), does elicit the error however.

src/time_zone_info.cc
```diff
@@ -470,7 +470,7 @@ bool TimeZoneInfo::Load(const std::string& name, ZoneInfoSource* zip) {
   // Determine the before-first-transition type.
   default_transition_type_ = 0;
   if (seen_type_0 && hdr.timecnt != 0) {
-    std::uint_fast8_t index = 0;
+    std::int_fast8_t index = 0;
```
devbww: I still don't understand why this should be changed.  Indeed, without it (the status quo) -Wsign-compare compiles cleanly, whereas with the change to int_fast8_t I see warnings firing.

So, I'm going to close this PR.  Please followup if you believe there are still issues.  Thanks.

PR #51, opened 2017-09-22
Declaring kDigits static to avoid a codegen bug in VS2017.

mwinterb:
The issue should be fixed shortly, but until then.
https://developercommunity.visualstudio.com/comments/114235/view.html

### DIFF
--- a/src/time_zone_fixed.cc
+++ b/src/time_zone_fixed.cc
@@ -28,7 +28,7 @@ namespace {
 const char kFixedOffsetPrefix[] = "Fixed/";
 
 int Parse02d(const char* p) {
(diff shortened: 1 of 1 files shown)


mwinterb:
This seemed too noisy for the commit message, but clang generates the same code regardless of static or not. gcc generates different code: with static references a global array, without static, gcc builds up the array from several immediates. I'm not in a position to to judge which is better, but since Parse02d isn't really called on a hotpath, the differences are hopefully irrelevant.
https://godbolt.org/g/dkGFm7

devjgm:
Looks good. Thanks for the fix here and the bug report to Microsoft.

PR #54, opened 2017-10-17
add CMake support

Sarcasm:
This change should make it easier to use for https://github.com/abseil/abseil-cpp/
were CMake work is in progress,
and for vcpkg which added a definition:
- https://github.com/Microsoft/vcpkg/blob/master/ports/cctz/CMakeLists.txt

This should also help with the build on Windows,
building the library is supported.
The examples do not run very well but I think this is some issues with
missing timezone files.

The Makefile is kept but the CMake files should produce the same output,
with the addition of CMake package config files.


Sarcasm:
FYI, I have rebased the conflict on the README.md.

devbww:
Hi Guillaume,

We are most definitely interested in this change (and thanks for merging in the conflicting commit).

At the moment we are considering whether to try to include compiled tzdata within cctz.  This would makes things easier for Windows users and also make the tests self-contained (by removing any dependency on the local data).  But it would also mean that the structure of the build rules might change.

So, we'd like to hold off for a few days.  If we decide to leave things the same, we will happily pull your change then.  If things do change however, we would appreciate your help in updating the CMake support.

How does that sound?

thanks,
Bradley

Sarcasm:
Sounds good, I can understand how this would help the out of the box experience, at least for Windows users.

JonathanDCohen:
Any progress here?  Abseil CMake support will land very soon, this will be a big usability upgrade

Sarcasm:
From my side, I'm waiting to be notified if I need to change something regarding the bundling of the tzdata.

I saw a commit which imported the tzdata, so I updated my patch to use them in the tests:

```cmake
  # tests runs on testdata
  set_property(
    TEST
      civil_time_test
      time_zone_format_test
      time_zone_lookup_test
    PROPERTY
      ENVIRONMENT "TZDIR=${CMAKE_CURRENT_SOURCE_DIR}/testdata/zoneinfo"
    )
```

-- https://github.com/Sarcasm/cctz/blob/ed704b1e5af08ccee8d9eccd98b5829c8f361699/CMakeLists.txt#L138-L146

It seemed to work on Linux, I did not test on Windows.

For what its worth, I don't think the bundling should change much to the way the actual targets are defined.
So maybe this could be done in 2 commits, first a commit in its actual form, maybe with modifications if review requires it, then bundle the tzdata.

Sarcasm:
I have updated the patch to take the new zone_info_source.cc file from https://github.com/google/cctz/commit/4f9776a310f4952454636363def82c2bf6641d5f into account.

I also declared an `ALIAS` library `cctz::cctz`, so people using `add_subdirectory(cctz)` can use the same name for the library as those using `cctz` from an install tree.

The tests and examples use this new target, so people looking at the CMakeLists.txt can take it as an example.

JonathanDCohen:
"CMake Error at cctz/CMakeLists.txt:107 (find_package):
  By not providing "Findbenchmark.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "benchmark", but CMake did not find one.

  Could not find a package configuration file provided by "benchmark" with
  any of the following names:

    benchmarkConfig.cmake
    benchmark-config.cmake

  Add the installation prefix of "benchmark" to CMAKE_PREFIX_PATH or set
  "benchmark_DIR" to a directory containing one of the above files.  If
  "benchmark" provides a separate development package or SDK, be sure it has
  been installed."

I got this warning trying to use this PR to build Abseil

Sarcasm:
Yes that's expected.
If you just want to build the library, like the Makefile do, you can use `cmake -DBUILD_TESTING=OFF ...`.

Otherwise, you have to make Benchmark and the other dependencies available to CMake.

It's up to you to decide:
- build benchmark, install it in a directory, and use `-DCMAKE_PREFIX_PATH=<benchmark/install/dir>`
- add it with `add_subdirectory()`, and hide override `find_package()` as some people seems to suggest https://www.youtube.com/watch?v=bsXLMQ6WgIk&t=52m36s
- use one of the many tool capable of managing CMake dependencies

If you need to force the C++ standard to a specific version, you can use `-DCMAKE_CXX_STANDARD=[11|14|17]`, default is C++11.

devbww:
Sorry ... I meant to get back to you yesterday.

It looks like you've already covered all the points I wanted to mention: no dependencies on new testdata/zoneinfo, but addition of new src/zone_source_info.cc file.

I don't understand the benchmark configuration stuff, but it would be nice if it "just worked."  Is that feasible?

Sarcasm:
> Is that feasible?

It is *technically possible*, yes.
However, CMake is not a package manager, so there is no perfect/right/clean way to do this, IMHO.
In order to keep the CMake files clean, I preferred to leave this out.
The code and the examples can be built without.

Maybe something not too intrusive is to document the multiple operations in the README?
I could also attempt to do this in a `.travis.yml`, even if the Github project does not uses it in the end, it could serve as an example.

What do you think?

CMakeLists.txt
```diff
@@ -0,0 +1,152 @@
...
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+  add_compile_options(-Wall)
+endif()
+
+# if a specific standard is required, this can be set from the command line:
+#     cmake -DCMAKE_CXX_STANDARD=[11|14|17]
+if (NOT CMAKE_CXX_STANDARD)
+  # C++11 is the minimum version required
+  set(CMAKE_CXX_STANDARD 11)
```
Sydius: Would it be better to set the property on the target (also supported as of v3.1, which is the only time this takes effect anyway) so that it is transitive? I'm assuming the public headers also require C++11.

CMakeLists.txt
```diff
@@ -0,0 +1,152 @@
...
+endif()
+
+# CMAKE_CXX_STANDARD is only supported since CMake >= 3.1
+if (CMAKE_VERSION VERSION_LESS "3.1")
+  if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    add_compile_options(-std=c++${CMAKE_CXX_STANDARD})
+  endif ()
+endif()
+
+# Setup the flag to enable colored diagnostics if any.
```
Sydius: It seems strange to me to include color diagnostics in a project like this, which is almost certainly going to have its own settings.

CMakeLists.txt
Sarcasm: Ah! Thanks for making me look into this.

It is my understanding that CMAKE_CXX_STANDARD is more targeted to toolchain files than for the projects directly<sup>[1](#footnote-1)</sup>, so this would be an improvement.

I thought the way to do this per-target was:

```cmake
target_compile_features(cctz PUBLIC cxx_std_11)
```

But this is only available since CMake 3.8.
However, you right, because an alternative syntax exists:

```cmake
set_property(TARGET cctz PROPERTY CXX_STANDARD 11)
```

Will make the change once I find something that can work ok with CMake < 3.1,
and if possible respects CMAKE_CXX_STANDARD.

---

<a name="footnote-1">1</a>: https://gitlab.kitware.com/cmake/cmake/issues/17146#note_299405

CMakeLists.txt
Sarcasm: I think I will remove this code as this is "not standard", and there is currently no way to enable or disable it.
I need to find a less intrusive way to do this.

However, I don't think it should be an issue in practice for projects including cctz in another projects, as `add_compile_options()` should not affect them.

JonathanDCohen:
@Sarcasm ok this all sounds good.

A couple random tohguhts without looking at this PR too deeply:

 * We should hold off on install() rules until they come to absl.  We're leaning towards accepting https://github.com/abseil/abseil-cpp/issues/38 in some capacity or another to make Abseil more usable through common build tooling, and when that happens cctz should look similar

 * If Windows is a problem, let's land this for Linux / OSX and then do Windows support in a follow-up PR.

 * Is there a way such that cctz's dep on benchmark can be made transparent to people who want to use Abseil without having to add extra build flags?  If no, then the README for Abseil and cctz need to have directions on how to mitigate that error.

CMakeLists.txt
Sarcasm: Just a follow up.

Using `-DCMAKE_CXX_ARGS=-fcolor-diagnostics` is a non-intrusive way to do this, so one can get colors with Ninja using the following `cmake` invocation:

    cmake -G Ninja -DCMAKE_CXX_ARGS=-fcolor-diagnostics ...

Can also be done with `CXXFLAGS=-fcolor-diagnostics cmake ...`.

Sarcasm:
@JonathanDCohen 

> We should hold off on install() rules until they come to absl. We're leaning towards accepting abseil/abseil-cpp#38 in some capacity or another to make Abseil more usable through common build tooling, and when that happens cctz should look similar

Done.
I'd like to complain a bit, because I'm sad to remove them.
I added install targets  for 2 reasons:
1. the Makefile currently provides `make install`
2. I didn't know `cctz` had the same same ABI compatibility issues as Abseil, it looked like a "more traditional" library to me

Not sure to understand why cctz "should look similar" to abseil.
I tried to make cctz be similar to a "standard" CMake project, with no idiosyncrasies.

Anyway, it's not because they are removed today that they won't see the light of day some day, one can still have hope. :)

> Is there a way such that cctz's dep on benchmark can be made transparent to people who want to use Abseil without having to add extra build flags? If no, then the README for Abseil and cctz need to have directions on how to mitigate that error.

Sorry, I don't really understand the question.

However, I have implemented a second CMake project, which builds a workspace for cctz and its dependencies.
The idea is that, if you want to work on cctz, or test the build you can use this project.
But, if you want to use cctz to integrate it into your own workspace, then you need to create a similar workspace for your project.

You can give it a try, it build Benchmark and googletest/googlemock from source.

- https://github.com/Sarcasm/cctz/blob/793f6327f5489bea8931d4fb41f2e3e4a7b46fc4/cmake-workspace/README.md

I could make the same sort of "cmake-workspace" project for Abseil, however, without an install target for `cctz` I'm not sure how to proceed.

I invite you to give it a try, but I'm ok removing it from the PR if it's considered "out of scope".

JonathanDCohen:
@Sarcasm there will likelybe install() support for Abseil, it just needs to be worked out how it's provided, and cctz I would think should follow suit.

We're definitely talking about a workspace project for Abseil and its transitive deps.  Let's make a separate Issue for that on the abseil-cpp project and figure out what that should look like.

CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+  src/time_zone_posix.h
+  src/tzfile.h
+  src/zone_info_source.cc
+  )
+cctz_target_set_cxx_standard(cctz)
+add_library(cctz::cctz ALIAS cctz)
+
+target_include_directories(cctz PUBLIC PUBLIC include)
+
+# not (yet) supported on Windows because of getopt.h
```
Sydius: Use CHECK_INCLUDE_FILE instead of MSVC?

CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+project(cctz)
+
+set(CMAKE_MODULE_PATH
+  ${PROJECT_SOURCE_DIR}/cmake
+  ${PROJECT_SOURCE_DIR}/cmake/modules
+  ${CMAKE_MODULE_PATH})
+
+include(CheckCXXCompilerFlag)
+include(CTest)
+include(GNUInstallDirs)
```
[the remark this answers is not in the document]
Sarcasm: Ah thanks, also removed `include(CheckCXXCompilerFlag)`.

CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+
+option(BUILD_EXAMPLES "Whether or not to build examples" ON)
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+  add_compile_options(-Wall)
+endif()
+
+function(cctz_target_set_cxx_standard target)
+  set(cctz_cxx_standard 11)
```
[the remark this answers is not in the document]
Sarcasm: I will ignore, not because I feel strong about it, but because I have a reasoning which you might consider okay once I share it with you.

I like to name my project/directory variables uppercase, because the scope is somewhat global, inherited from directory to directory.
Here the variable is in the function local scope, very localized, so I feel like the name does not need to be so distinctive.

Checking a few CMake modules, not too dated, such as FindThreads.cmake, FindGTest.cmake, FindProtobuf.cmake, FindPythonLibs.cmake, I would say it is somewhat similar.
Eventually they also add an underscore prefix.
I guess I could use something like `_cxx_standard` if I wanted to copy the style.

CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+  src/time_zone_lookup.cc
+  src/time_zone_posix.cc
+  src/time_zone_posix.h
+  src/tzfile.h
+  src/zone_info_source.cc
+  )
+cctz_target_set_cxx_standard(cctz)
+add_library(cctz::cctz ALIAS cctz)
+
+target_include_directories(cctz PUBLIC PUBLIC include)
```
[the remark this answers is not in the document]
Sarcasm: Ah thanks for catching that error.
I don't feel like trying `PRIVATE` + `SYSTEM PUBLIC`, I'm pretty sure that if found via an installed config, CMake would consider it a SYSTEM header automatically.
And I don't know for projects that includes the subdirectory directly, but I don't think we should special case for it, if they include the source directly, they might as well have our warnings.

CMakeLists.txt
Sarcasm: Done with `check_cxx_symbol_exists`, I assume it's theoretically possible for a system to have getopt.h without getopt_long, which is a GNU extension.
The CMake code is slightly more complicated with this change, but I guess the right thing to do would be to not use `getopt_long`.

Sarcasm:
@JonathanDCohen 

> We're definitely talking about a workspace project for Abseil and its transitive deps.
> Let's make a separate Issue for that on the abseil-cpp project and figure out what that should look like.

Honestly, the current Abseil CMakeLists.txt makes this quite challenging, but it's worth discussing.

@Sydius 

> I'm not sure about the workspace thing because it's not a pattern I've personally seen before, but it seems ok at first glance.

I did not see this pattern much either, and I'm not sure CMake is an ideal tool to manage a workspace, but right now I'm thinking this method is somewhat ok. It keeps the project clean, and you can have a reasonably simple workspace to work with.

I totally agree with the inheritance issue you mention.

CMakeLists.txt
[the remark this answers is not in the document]
Sydius: Sounds reasonable.

CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
...
+check_cxx_symbol_exists(getopt_long getopt.h HAVE_GETOPT_LONG)
+if (HAVE_GETOPT_LONG)
+  add_executable(time_tool src/time_tool.cc)
+  cctz_target_set_cxx_standard(time_tool)
+  target_link_libraries(time_tool cctz::cctz)
+elseif(show_time_tool_msg)
+    message(STATUS "Disable time_tool as getopt_long is not available")
+endif()
+
+if (BUILD_EXAMPLES)
```
Sydius: Can BUILD_EXAMPLES be CCTZ_BUILD_EXAMPLES and BUILD_TESTING be CCTZ_BUILD_TESTING?

My project pulls cctz in via add_subdirectory() (to avoid the external projects inheritance mess) and I would hate to have to set something so generic sounding one level up.

Sarcasm:
> I would leave the "workspace" part out and put that as a separate thing, personally, but I don't feel strongly about it.

Removed.

CMakeLists.txt
Sarcasm: I would prefer not to.
I think this is the expected, simplest, way to use options.
`BUILD_TESTING` and other options like `BUILD_SHARED_LIBS`, are actually declared by CMake.
I'm seeing quite a few project namespacing these options to specifically for CMakeLists.txt that uses `add_subdirectory(cctz)`.
But I don't think `add_subdirectory()` needs any special treatment, I have actually realized this when someone on the Abseil CMake PR made me realize how safe `add_subdirectory` was.

About a week ago I think, I saw someone talk about this exact problem on the cppclang Slack, in the \#cmake channel, and the exposed solution is both simple and clean IMHO.

The super-project can just wrap the `add_subdirectory` call with the appropriate value for `BUILD_TESTING` and friends.
It can decide to:
- disable it or enable it
- to have a specific option `CCTZ_BUILD_EXAMPLES` 
- or a grouped option like `DEPS_BUILD_EXAMPLES`
- do anything else, even nothing

I haven't actually tested, so I hope this is correct...let me know.

Sydius:
OK, I tested this just now and all works pretty well. LGTM.

Sarcasm:
1. Do you want me to squash the commits into one?
2. Should I add myself in a CODEOWNER for the `cmake/` directory and the `CMakeLists.txt`?
   If I get notified about the CMake changes I can get an eye to the change.
   - https://help.github.com/articles/about-codeowners/

devbww:
Hey, I just committed a change to remove <getopt.h> from src/time_tool.cc, which should remove the need for special HAVE_GETOPT_LONG treatment in CMakeLists.txt.  I would happily merge a pull request that did that.  Thanks.

Sarcasm:
I have opened a PR:
- https://github.com/google/cctz/pull/64

PR #61, opened 2017-11-11
Provide a means of disabling testing

Sydius:
When CCTZ is pulled into another CMake project using
add_subdirectory(), there's no good way (that I could find) to disable
the building of tests without disabling BUILD_TESTING globally for the
project (since it's an 'option' in the cache, not specific to the
directory).

On Linux, set(BUILD_TESTING OFF) worked, but this doesn't seem to work
on Windows.

This change introduces CCTZ_BUILD_TESTING as an additional way of
turning it off. Also renamed BUILD_EXAMPLES to CCTZ_BUILD_EXAMPLES for
consistency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,11 @@ set(CMAKE_MODULE_PATH
 include(CheckCXXSymbolExists)
 include(CTest)
 
-option(BUILD_EXAMPLES "Whether or not to build examples" ON)
+option(CCTZ_BUILD_EXAMPLES "Whether or not to build examples" ON)
+
+# CMake does not seem to provide a means of disabling testing for a subproject,
+# so providing an option similar to BUILD_TESTING specific to the project.
+option(CCTZ_BUILD_TESTING "Whether or not to build tests" ON)
 
 # Starting from CMake >= 3.1, if a specific standard is required,
 # it can be set from the command line with:
@@ -66,11 +70,11 @@ elseif(show_time_tool_msg)
     message(STATUS "Disable time_tool as getopt_long is not available")
 endif()
 
-if (BUILD_EXAMPLES)
+if (CCTZ_BUILD_EXAMPLES)
   add_subdirectory(examples)
 endif()
 
-if (BUILD_TESTING)
+if (CCTZ_BUILD_TESTING AND BUILD_TESTING)
   find_package(benchmark REQUIRED)
   find_package(GMock REQUIRED)
   find_package(GTest REQUIRED)


Sydius:
@Sarcasm -- PTAL -- I couldn't get it to work because of the interactions with the cache. Open to alternatives if there's a better way.

Sarcasm:
While this can be worked around in cctz, I'm wondering if there is not a non-intrusive way to do this.
People who want to use `add_subdirectory()` should strive for doing it without modifying the subprojects.

Here is an example:

```cmake
cmake_minimum_required(VERSION 3.3)

project(cctztest)

include(CTest)

macro(option option_name)
  if (${option_name} IN_LIST BUILD_TESTING)
    message(STATUS "BUILD_TESTING for cctz detected, override")
    set(BUILD_TESTING OFF)
  endif()
  _option(${ARGV})
endmacro()

add_subdirectory(cctz)
```

This can probably be made cleaner by wrapping it into something like:

```cmake
option(CCTZ_BUILD_EXAMPLES "Whether or not to build cctz examples" OFF)

myproject_add_subproject_directory(cctz
 OVERRIDE_OPTIONS
   BUILD_TESTING OFF
   BUILD_EXAMPLES ${CCTZ_BUILD_EXAMPLES}
   BUILD_SHARED_LIBS ${CCTZ_BUILD_SHARED_LIBS}
)
```

This approach is similar to the `find_package()` override of the "effective CMake" talk:
- https://www.slideshare.net/DanielPfeifer1/effective-cmake (slide 63)

Another thought I did not explore, is the file `CMAKE_PROJECT_cctz_INCLUDE`:
- https://cmake.org/cmake/help/v3.9/variable/CMAKE_PROJECT_PROJECT-NAME_INCLUDE.html

I'd like to see if the people on the cpplang \#cmake channel can come up with good solutions as well.

Sydius:
Hmm. I've been having trouble getting it to work for some reason. Insofar as I can tell, BUILD_TESTING is set in the cache and using 'set' before 'add_subdirectory' doesn't override it on Windows. I can disable it by force-updating the cache, but then it's globally disabled for my entire project and not just for the sub-directory.

Sarcasm:
> Insofar as I can tell, BUILD_TESTING is set in the cache
> and using 'set' before 'add_subdirectory' doesn't override it on Windows.

Indeed, set does not override when the variable is in cache,
and that is why I did not propose this in my solution.
I noticed this on Linux, so I'm not sure if there is any behavior specific on Windows,
I assume it's just the CMake behavior, disregarding any platform specificities.

> I can disable it by force-updating the cache,
> but then it's globally disabled for my entire project
> and not just for the sub-directory.

Yes, you would have to disable it before `add_subdirectory(cctz)`
and then reenable it after.
The issue with this solution is that, if the `add_subdirectory()` fails,
then CMake exits before you have the opportunity to restore the value in the cache.
That's why I did not propose this solution either.

Out of curiosity, did you try my example?
Overriding macro before including cctz?

```cmake
# define top level options before the redefinition of option()
option(BUILD_EXAMPLES "Whether or not to build the main examples" ON)

# override options for subdirectories
# here this is not very generic,
# but it can be wrapped into something cleaner/less global
macro(option option_name)
  if (${option_name} IN_LIST BUILD_TESTING)
    message(STATUS "BUILD_TESTING for cctz detected, override")
    set(BUILD_TESTING OFF)
  endif()
  _option(${ARGV})
endmacro()

# add subdirectories here, each of them should have BUILD_TESTING disabled
add_subdirectory(cctz)
```

The macro replaces `option(BUILD_TESTING ...)` by `set(BUILD_TESTING ...)`,
which worked well in my mininaml test.

Sydius:
So I thought I had tried something equivalent to your example, but I see now that I didn't. Apologies for leaving this open for so long; been too busy at work and my home computer's motherboard died on me.

I'll try to find some time to test out what you suggested and see if I can avoid editing the project.

Sarcasm:
FWIW, I asked on the cpplang \#cmake channel if they had better ways to handle this issue.
The answer was basically that this is a CMake anti-pattern, and was redirected to:
- https://www.reddit.com/r/cpp/comments/65iyxv/embracing_modern_cmake/dgjamhk/

An excerpt:

> The way I think of this is to consider some C++:
>
>    ```cpp
>    struct Buildsystem
>    {
>        std::string globalSettings;
>        std::vector<std::string> libraries;
>        // ...
>    
>    
>        void consume(Buildsystem const& other)
>        {
>            // Oops, overwriting things!
>            globalSettings = other.globalSettings;
>    
>            // Oops, libraries is not unique anymore!
>            libraries.insert(libraries.end(), other.libraries.begin(), other.libraries.end());
>        }
>    };
>    ```
>
> The consume method will wreak havok on your data because it doesn't know any better.
> That's what happens when you `add_subdirectory()` random projects and I'm very against doing so.

Wanted to share this, because I agree with it, and it's coming from a CMake contributor, one who implemented many of the "modern CMake" features.

Anyway, if my workaround works I would be glad.

Sydius:
It works if I add an option(BUILD_TESTING "" ON) next to the existing BUILD_EXAMPLES in the cctz CMakeLists.txt. Without that, though, it doesn't work on the first invocation of CMake for whatever reason (but it does work on subsequent runs).

CMakeLists.txt
```diff
@@ -66,11 +70,11 @@ elseif(show_time_tool_msg)
     message(STATUS "Disable time_tool as getopt_long is not available")
 endif()
 
-if (BUILD_EXAMPLES)
+if (CCTZ_BUILD_EXAMPLES)
   add_subdirectory(examples)
 endif()
 
-if (BUILD_TESTING)
+if (CCTZ_BUILD_TESTING AND BUILD_TESTING)
```
Sarcasm: I think it would be better to use only CCTZ_BUILD_TESTING, it's better if there is only one way to do things.

Sarcasm:
> It works if I add an option(BUILD_TESTING "" ON) next to the existing BUILD_EXAMPLES in the cctz CMakeLists.txt.

It's weird, because what we currently have in cctz is:

https://github.com/google/cctz/blob/ecb90a9bab71187f691b204affc3e688eb8e4455/CMakeLists.txt#L11-L13


and the first line of code in `include(CTest)`, is actually just declaring BUILD_TESTING:
- https://github.com/Kitware/CMake/blob/8b5ae1c1d310fc6eb13b2cde5712b4e319f01f3e/Modules/CTest.cmake#L54

Sydius:
Yeah, I can't really explain it. I added messages that print out when the macro is invoked and what is passed to it. The first and second time both invoke it and both have identical input, but the if statement only triggers on the second. I'm sure I'm just doing something silly, but I couldn't figure it out after about 30 minutes of fussing with it. :/

At any rate, I'm OK with abandoning this if it's not useful to others. I can just carry the patch locally; just uploaded it thinking more people would run into the same issue.

CMakeLists.txt
devbww: Any action here, or has this been abandoned?  I'm happy to merge if the CMake experts agree.

Sarcasm:
@devbww: Since nobody else asked for it yet maybe it can be closed, it will always be possible to reopen later if the needs arise.

Also, I'm not sure, but it's possible that updating to the latest CMake (3.13), fixes the initial issue of `set` not overriding the option.

See this part of the release notes:

> The `option()` command now honors an existing normal variable of the same name and does nothing instead of possibly creating a cache entry (or setting its type) and removing the normal variable. See policy CMP0077.
> -- https://cmake.org/cmake/help/v3.13/release/3.13.html?highlight=cache#other-changes

However, after some attempts, I wasn't able to reproduce the initial issue, so I cannot tell if the release fixes it.

devbww:
OK, thanks @Sarcasm.  Either way it sounds best to wait to see if problems (re)arise.  So I'll close for now.  Thanks again.

PR #68, opened 2018-02-09
Add install headers and library with CMake

vozbu:
This is a very simple realization of installation in CMake. So-called "works for me". Tests and examples are not installed.

I think it might be handy because recently I used manually copying files to destination because of lack of installation support in CMakeLists.txt. If only you don't have some objections against having installation here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,8 @@ if (BUILD_TESTING)
   )
 endif()
 
+install (DIRECTORY include/cctz DESTINATION include)
+
 # Starting from CMake >= 3.1, if a specific standard is required,
 # it can be set from the command line with:
 #     cmake -DCMAKE_CXX_STANDARD=[11|14|17]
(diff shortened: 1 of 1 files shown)


vozbu:
Oh, CLA. I wasn't very attentive sorry.

vozbu:
I signed it!

Sarcasm:
Just for the record, this was initially proposed in the PR that introduced CMake, but it had to be removed because of the Abseil policy regarding installation:

> We should hold off on install() rules until they come to absl.
> We're leaning towards accepting abseil/abseil-cpp#38 in some capacity or another to make Abseil more usable through common build tooling, and when that happens cctz should look similar
> -- @JonathanDCohen, https://github.com/google/cctz/pull/54#issuecomment-341726882

The commit that removed the install targets is here:

- https://github.com/Sarcasm/cctz/commit/79c7c0ad5e488e405e3c70aca519311bad8804d6

It is more complete in that it installs the CMake config files so that users of cctz can just call `find_package(cctz REQUIRED)`.

vozbu:
Aha, I just missed that looking through CMakeLists.txt history... Obviously, your version was  much better. Probably, this pull request can be safely closed now.

devbww:
Thanks for keeping an eye on things Guillaume.  It is a shame that CCTZ is being constrained by Abseil, but probably best to wait until they get their CMake story straightened out.  In the meantime I hope that everyone can handle this downstream.

PR #74, opened 2018-04-25
Enable C++14 constexpr for more compilers

rongjiecomputer:
Many compilers can handle C++14 constexpr already, let's not restrict ourselves to clang only.

Visual Studio 2015 does not have `__cpp_constexpr` macro, but the compiler always defaults to C++14 mode, so it is enough to just check the compiler version.

Tested with Mingw64 GCC 7.1.0, Clang/MSVC (master) and Visual Studio 2017 15.5, all in C++14 mode.

### DIFF
--- a/include/cctz/civil_time_detail.h
+++ b/include/cctz/civil_time_detail.h
@@ -20,8 +20,9 @@
 #include <ostream>
 #include <type_traits>
 
-// Disable constexpr support unless we are using clang in C++14 mode.
-#if __clang__ && __cpp_constexpr >= 201304
+// Disable constexpr support unless we are in C++14 mode.
+#if (defined(__cpp_constexpr) && __cpp_constexpr >= 201304) || \
+    (defined(_MSC_VER) && _MSC_VER >= 1900)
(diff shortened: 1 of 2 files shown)


devbww:
I tried it but saw errors like ...

`include/cctz/civil_time_detail.h(92): error C3250: 'yi': declaration is not allowed in 'constexpr' function body`

It is all behind too many layers for me to tell which _MSC_VER I was getting, however.  Are we sure "1900" is enough?

rongjiecomputer:
@devbww Ops, I double-checked the MSVC docs again. C++14 relaxed constexpr is available [only since Visual Studio "15" Preview 5 (fifth preview before Visual Studio 2017)][0]. Visual Studio 2015 does not have it.

`_MSC_VER` numbers come from https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering

[0]: https://blogs.msdn.microsoft.com/vcblog/2016/10/05/visual-studio-15-preview-5-now-available/

PR #76, opened 2018-05-30
Update WORKSPACE; google/benchmark has bazel support

astrelni:
https://github.com/google/benchmark/blob/master/BUILD.bazel

Ran bazel run -c opt :benchmarks to check functionality.


astrelni:
Oops, just noticed the http_archive name needs to be updated to match the benchmark WORKSPACE, one moment.

PR #81, opened 2018-09-29
Mingw gcc doesn't define _MSC_VER, use _WIN32 instead.

sighingnow:


devbww:
Might not it be nicer, if we want HAS_STRPTIME=0 to be the default for Mingw, to add some Mingw-specific case instead of overloading _WIN32?  Or is "Windows" really the best indication of a missing strptime()?  Thanks.

sighingnow:
Now the function `strptime` is checked in CMake script. I think it is a better solution for this trouble than previous hard-coded approach.

rongjiecomputer:
This will break Bazel user on Windows as Bazel does not have an equivalent for CMake's `check_function_exists`.

devbww:
There was a recent change here, [https://github.com/google/cctz/pull/87](https://github.com/google/cctz/pull/87), that affected MinGW behavior.  Please make another PR if that is insufficient for you.  Thanks.

PR #83, opened 2018-10-04
Add CMake install rules

### LINKED ISSUE #82: Missing CMake install rules
Currently cctz doesn't provide CMake install rule which is a pain for CMake users to integrate it.

On the same subject:

* There was a proposal in the PR #54, but you ask for removal cf https://github.com/google/cctz/pull/54/commits/79c7c0ad5e488e405e3c70aca519311bad8804d6
* Distro maitainers have to maintain patchs to do it https://salsa.debian.org/science-team/cctz/blob/d335bb4309d960b88e5f9024983f6f934a6db366/debian/patches/0001-Compile-shared-lib-and-install-it.patch
* Same for vcpkg, microsoft dev already provide an install rule whithout waiting for you... https://github.com/Microsoft/vcpkg/blob/ab48a71c9af25041ab29a0d8e62a3cdf380638a5/ports/cctz/CMakeLists.txt#L21-L36
* On the other hands Makefile **already** provide install rules cf https://github.com/google/cctz/blob/a2acba2966660a9bfe9ab66efcfad573d91f56b6/Makefile#L85-L97
* For https://github.com/google/or-tools we are currently migrating from "custom" base to abseil -> I have to patch cctz and maintain it cf https://github.com/google/or-tools/blob/abseil/patches/cctz-master.patch
EDIT: Abseil-cpp now has a cctz version embedded...

Mizux:
Add install rules to cctz to fix #82 issue

inspired by the revert commit in #54

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,13 +58,14 @@ function(cctz_target_set_cxx_standard target)
   endif()
 endfunction()
 
-set(_HDRS
+set(CCTZ_HDRS
   include/cctz/time_zone.h
   include/cctz/civil_time_detail.h
   include/cctz/zone_info_source.h
   include/cctz/civil_time.h
   )
-set(_SRCS
+add_library(cctz "")
+target_sources(cctz PRIVATE
   src/civil_time_detail.cc
   src/time_zone_fixed.cc
   src/time_zone_fixed.h
@@ -82,15 +83,15 @@ set(_SRCS
   src/time_zone_posix.h
   src/tzfile.h
   src/zone_info_source.cc
+  ${CCTZ_HDRS}
   )
-add_library(cctz ${_SRCS} ${_HDRS})
 cctz_target_set_cxx_standard(cctz)
 target_include_directories(cctz PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include>
   )
 set_target_properties(cctz PROPERTIES
-  PUBLIC_HEADER "${_HDRS}"
+  PUBLIC_HEADER "${CCTZ_HDRS}"
   )
 add_library(cctz::cctz ALIAS cctz)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,9 @@ set(_SRCS
   )
 add_library(cctz ${_SRCS} ${_HDRS})
 cctz_target_set_cxx_standard(cctz)
+set_target_properties(cctz PROPERTIES
+  PUBLIC_HEADER "${_HDRS}"
+  )
 add_library(cctz::cctz ALIAS cctz)
 
 target_include_directories(cctz PUBLIC include)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,12 +85,14 @@ set(_SRCS
   )
 add_library(cctz ${_SRCS} ${_HDRS})
 cctz_target_set_cxx_standard(cctz)
+target_include_directories(cctz PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include>
+  )
 set_target_properties(cctz PROPERTIES
   PUBLIC_HEADER "${_HDRS}"
   )
 add_library(cctz::cctz ALIAS cctz)
-
-target_include_directories(cctz PUBLIC include)
 
 if (BUILD_TOOLS)
   add_executable(time_tool src/time_tool.cc)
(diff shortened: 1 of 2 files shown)


Mizux:
Bazel build is broken in kokoro https://sponge.corp.google.com/target?id=34c97a92-34d7-4fbd-9803-dce22e5ebdfc&target=cctz%2Fgithub%2Fgcp_windows%2Fpresubmit&searchFor=

You don't even test CMake based build :disappointed:

devbww:
@Sarcasm, @sigmoidal, or @JonathanDCohen:  Any comments/advice.  Thanks.

CMakeLists.txt
```diff
@@ -58,7 +58,13 @@ function(cctz_target_set_cxx_standard target)
...
 endfunction()
 
-add_library(cctz
+set(_HDRS
+  include/cctz/time_zone.h
+  include/cctz/civil_time_detail.h
+  include/cctz/zone_info_source.h
+  include/cctz/civil_time.h
+  )
+set(_SRCS
```
Sarcasm: Note: in my initial CMake patch, I purposefully did not make such variables, because I consider it not "Modern CMake", see Stephen Kelly talk:
- Embracing Modern CMake: https://steveire.wordpress.com/2017/11/05/embracing-modern-cmake/
- Slides 49-52, "Reliance on variables": http://www.steveire.com/embracing-modern-cmake.pdf

Mizux: agree with directly adding sources in `add_library()` command directly or using `target_sources()`. I was lazy on this (often use GLOB recurse and no incremental build but containerized build from scratch).
I would like to keep headers also (didn't see any reference about it in Modern CMake) so IDE display them in there solution (while it is not strictly needed for the compilation)

Mizux: 1) [PUBLIC_HEADER](https://cmake.org/cmake/help/latest/prop_tgt/PUBLIC_HEADER.html) is also working outside macOS, look at the last sentence:

> On non-Apple platforms these headers may be installed using the PUBLIC_HEADER option to the install(TARGETS) command.

Thus it avoid the `install(FILES` and keep you in the Modern CMake way (IMHO).
This is also mentioned is the recent CMake documentation for [`install(TARGETS ...)`](https://cmake.org/cmake/help/latest/command/install.html#targets)

> PUBLIC_HEADER
Any PUBLIC_HEADER files associated with a library are installed in the destination specified by the PUBLIC_HEADER argument on non-Apple platforms.

I still have concern if we'll get a 'target_headers()' or similar in CMake later
note: no answer from kitware https://cmake.org/pipermail/cmake/2017-March/065126.html

1bis) if you set headers as `PUBLIC` target sources they'll also appear on any target linking against this target (e.g. executable target B will show library A headers in its source list)
I'll try to take screenshot on this...

2) **QtCreator** and **VisualStudio 2017** both support as built-in cmake project and will display headers if they are listed in the target [SOURCES](https://cmake.org/cmake/help/latest/prop_tgt/SOURCES.html)
note: need to try/confirm Eclipse and CodeBlocks...

2bis) This is not working for INTERFACE currently (at least on qtcreator)
https://bugreports.qt.io/browse/QTCREATORBUG-15677
https://gitlab.kitware.com/cmake/cmake/merge_requests/2282

Sarcasm: So, do IDEs make use of `PUBLIC_HEADER`, or only `SOURCES`? Because I'm not advocating to remove the headers from the CMakeLists.txt, I'm just wondering what is the benefits of specifying them as `PUBLIC_HEADER`, over just listing them in the `add_library`.

In my initial patch I did not use `install(FILES` , but `install(DIRECTORY`, since all public headers are in the same directory.

I could see the benefit of `PUBLIC_HEADER` approach when using a "merged header placement" source layout:
- https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/spec/data/spec.bs#src.header-placement.merged

Anyway, I think it would look slightly cleaner, but equally correct so I don't really mind.

CMakeLists.txt
```diff
@@ -58,7 +58,14 @@ function(cctz_target_set_cxx_standard target)
...
 endfunction()
 
-add_library(cctz
+set(CCTZ_HDRS
+  include/cctz/time_zone.h
+  include/cctz/civil_time_detail.h
+  include/cctz/zone_info_source.h
+  include/cctz/civil_time.h
+  )
+add_library(cctz "")
```
Sarcasm: cctz supports CMake 2.8.12 I believe, and `target_sources`, with such syntax, is a recent addition IIRC.

Also, what is the benefit for cctz of removing the sources from the `add_library` call?

Mizux: At first I wanted to use `target_source()` with `PUBLIC` for header...

CMakeLists.txt
Mizux: You can take a look at https://github.com/Mizux/plop/blob/master/README.md
sources PUBLIC need generator like include (otherwise you'll get cmake error)
ps: open image in new tab

Sarcasm: Thanks for the example.
I'm not sure `target_sources`'s `PUBLIC` is meant for public headers, I would be interested to know, but it feels off topic for this PR.

Mizux: Agree can be done later.
ps: if you get info on this topic I will be happy if you can share it with me.
pps: I also done few test in plop project to understand the effect of install rule on public syntax
TLDR:
```cmake
if(ENABLE_INSTALL)
  target_sources(B
    PRIVATE
    src/B.cc
    PUBLIC
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/b/B.h>
    $<INSTALL_INTERFACE:include/b/B.h>
    )
else()
  target_sources(B
    PRIVATE
    src/B.cc
    PUBLIC
    ${CMAKE_CURRENT_SOURCE_DIR}/include/b/B.h
    )
endif()
```
i.e. like target_include_directory() you need to use `(BUILD|INSTALL)_INTERFACE:` generator once you start to have install rules...

devbww:
Thanks all.

PR #84, opened 2018-10-10
[time_zone_posix.h] Replace invalid "nested anonymous struct" with valid C++

tkoeppe:
The original code is accepted by many compilers as a compatibility extension with C, but it is not valid C++, and it is trivial to fix.

### DIFF
--- a/src/time_zone_posix.h
+++ b/src/time_zone_posix.h
@@ -66,25 +66,35 @@ namespace cctz {
 // it would take us to another day, and perhaps week, or even month.
 struct PosixTransition {
   enum DateFormat { J, N, M };
-  struct {
-    DateFormat fmt;
-    union {
-      struct {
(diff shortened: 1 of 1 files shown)


devjgm:
Thanks, Thomas. Is there a compiler flag that will warn on this usage? Otherwise, how did you find it? Also, is there a cppreference link (or something) that talks a little more about nested unnamed structs?

tkoeppe:
Yes, `-pedantic` will do the trick: https://wandbox.org/permlink/1HRIHToEPaVrXpFM

I discovered it while trying to convince my employer to use correct C++, but they depend on Abseil and are thus blocked. :-)

tkoeppe:
> Also, is there a cppreference link (or something) that talks a little more about nested unnamed structs?

How about http://eel.is/c++draft/class.union.anon#1:

> Nested types [...] cannot be declared within an anonymous union

PR #87, opened 2018-12-07
Fixed build under MinGW.

excitoon:
In this code Windows previously meant Visual C++. This patch extends this to MinGW (both 32 and 64).

Kind regards,
Vladimir.


devbww:
Thanks.  Sorry for the delay.

PR #89, opened 2019-01-21
Solve GCC8 warnings

bstaletic:
These warnings were discovered when I tried to update abseil dependency my project depends on.

For more details take a look at https://github.com/abseil/abseil-cpp/pull/254

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -919,7 +919,7 @@ bool TimeZoneInfo::NextTransition(const time_point<seconds>& tp,
     ++begin;
   }
   std::int_fast64_t unix_time = ToUnixSeconds(tp);
-  const Transition target = { unix_time };
+  const Transition target = {unix_time, 0, civil_second(), civil_second()};
   const Transition* tr = std::upper_bound(begin, end, target,
                                           Transition::ByUnixTime());
   for (; tr != end; ++tr) {  // skip no-op transitions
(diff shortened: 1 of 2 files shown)


src/time_zone_info.cc
```diff
@@ -919,7 +919,7 @@ bool TimeZoneInfo::NextTransition(const time_point<seconds>& tp,
     ++begin;
   }
   std::int_fast64_t unix_time = ToUnixSeconds(tp);
-  const Transition target = { unix_time };
+  const Transition target = { unix_time, {}, {}, {} };
```
devbww: We might as well use the same form as line 819 (or change that to this).

src/time_zone_info.cc
```diff
@@ -954,7 +954,7 @@ bool TimeZoneInfo::PrevTransition(const time_point<seconds>& tp,
     }
     unix_time += 1;  // ceils
   }
-  const Transition target = { unix_time };
+  const Transition target = { unix_time, {}, {}, {} };
```
devbww: Ditto.

src/time_zone_libc.h
```diff
@@ -33,10 +33,10 @@ class TimeZoneLibC : public TimeZoneIf {
       const time_point<seconds>& tp) const override;
   time_zone::civil_lookup MakeTime(
       const civil_second& cs) const override;
-  bool NextTransition(const time_point<seconds>& tp,
-                      time_zone::civil_transition* trans) const override;
-  bool PrevTransition(const time_point<seconds>& tp,
-                      time_zone::civil_transition* trans) const override;
+  bool NextTransition(const time_point<seconds>&,
```
devbww: These edits seem unnecessary/undesirable.

bstaletic: Reverted.

src/time_zone_info.cc
bstaletic: Copied the line 819 here.

PR #90, opened 2019-02-01
Add pragma alternate name support for Windows ARM64

ThomsonTan:
The extra declaration `E` for `_M_IA64` and `_M_AMD64` than `_M_IX86` seems for `__ptr64` type, so it also applies to `_M_ARM64`.


devbww:
Thanks.

PR #91, opened 2019-02-15
First test if _MSC_VER is defined to avoid -Wundef

wantehchang:
CCTZ clients may need to compile their code with the -Wundef warning
flag. It will be helpful if CCTZ header files can be compiled without
the -Wundef warning.

How to avoid the -Wundef warning: If a macro may be undefined, we need
to first test whether the macro is defined before testing its value. We
can't rely on the C preprocessor rule that an undefined macro has the
value 0L.

### DIFF
--- a/include/cctz/civil_time_detail.h
+++ b/include/cctz/civil_time_detail.h
@@ -21,7 +21,7 @@
 #include <type_traits>
 
 // Disable constexpr support unless we are in C++14 mode.
-#if __cpp_constexpr >= 201304 || _MSC_VER >= 1910
+#if __cpp_constexpr >= 201304 || (defined(_MSC_VER) && _MSC_VER >= 1910)
 #define CONSTEXPR_D constexpr  // data
 #define CONSTEXPR_F constexpr  // function
 #define CONSTEXPR_M constexpr  // member


include/cctz/civil_time_detail.h
```diff
@@ -21,7 +21,7 @@
 #include <type_traits>
 
 // Disable constexpr support unless we are in C++14 mode.
-#if __cpp_constexpr >= 201304 || _MSC_VER >= 1910
+#if __cpp_constexpr >= 201304 || (defined(_MSC_VER) && _MSC_VER >= 1910)
```
devbww: I'm in two minds about this.

On the one hand, the pre-processor behavior is well defined and lots of code depends on it.  That is, I'm not a fan of `-Wundef`.

On the other hand, if we're going to be explicit, we should be so consistently.  For example, there is another instance of the same issue on that very line.

Thoughts?

devbww:
This has been addressed by #103.

wantehchang:
Hi Bradley: I am sorry I missed your reply on Feb. 20. I agree with what you said about the preprocessor behavior, but -Wundef is enabled in some of the projects that I work on. Thank you for addressing this in #103.

devbww:
No worries.  Thanks for bringing it to our attention.

PR #94, opened 2019-03-08
Make local_time_zone() work on iOS

lapfelix:
It looks like "/etc/localtime" can't be accessed on iOS, so `local_time_zone()` always falls back to UTC.

I used Core Foundation methods that are available on every Apple platform (since OS X 10.0, apparently).

Tested with iOS 10, 11 and 12.1.4

### DIFF
--- a/src/time_zone_lookup.cc
+++ b/src/time_zone_lookup.cc
@@ -20,6 +20,11 @@
 #include <dlfcn.h>
 #endif
 #endif
+
+#if defined(__APPLE__)
+#include <CoreFoundation/CFTimeZone.h>
+#endif
+
 #include <cstdlib>
 #include <cstring>
 #include <string>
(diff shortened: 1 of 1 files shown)


lapfelix:
I signed it!

src/time_zone_lookup.cc
```diff
@@ -128,6 +133,12 @@ time_zone local_time_zone() {
     if (__system_property_get("persist.sys.timezone", sysprop) > 0)
       tz_env = sysprop;
 #endif
+#if defined(__APPLE__)
+    CFTimeZoneRef system_time_zone = CFTimeZoneCopySystem();
```
devjgm: Does this `system_time_zone` need to be CFRelease()'d?

lapfelix: Ah, yes! I definitely need to release that `system_time_zone` since it's copied. I just pushed a new commit which addresses this and another memory management issue.

src/time_zone_lookup.cc
```diff
@@ -152,6 +162,9 @@ time_zone local_time_zone() {
   free(localtime_env);
   free(tz_env);
 #endif
+#if defined(__APPLE__)
```
devbww: Given that the conditional above is `#elif`, might as well make this one too.

lapfelix: Sure. Done.

PR #99, opened 2019-03-22
Remove use of CFStringGetCStringPtr

### LINKED ISSUE #98: Recent commit 805a025 may break non-English mac users
Hello, while I'm debugging my strange segmentation fault, I reached to the patch 805a025.
https://github.com/google/cctz/commit/805a025a896994ed507c9704cce15221bac6a632#diff-369c46982c3f44e7ed265e8576edfd4d

This patch introduced a call to `CFStringGetSystemEncoding()` mac api, but this api seems to be broken in our non-English mac machines (I confirmed in Japanese language). This function returns 0 (`kCFStringEncodingMacRoman`) for English users, but it returns 1 for Japanese users, which is an undefined value!
https://opensource.apple.com/source/CF/CF-299/String.subproj/CFString.h.auto.html

My mac version: 10.14.3
My xcode version: Version 10.0 (10A255)

devjgm:
`CFStringGetCStringPtr` might return nullptr, which causes `strdup` to crash. This PR now uses `CFStringGetString`, which writes to a local buffer.

Fixes #98

### DIFF
--- a/src/time_zone_lookup.cc
+++ b/src/time_zone_lookup.cc
@@ -126,8 +126,15 @@ time_zone local_time_zone() {
   _dupenv_s(&tz_env, nullptr, "TZ");
 #elif defined(__APPLE__)
   CFTimeZoneRef system_time_zone = CFTimeZoneCopyDefault();
-  CFStringRef tz_name = CFTimeZoneGetName(system_time_zone);
-  tz_env = strdup(CFStringGetCStringPtr(tz_name, CFStringGetSystemEncoding()));
+  if (CFStringRef tz_name = CFTimeZoneGetName(system_time_zone)) {
+    CFStringEncoding encoding = kCFStringEncodingUTF8;
+    CFIndex length = CFStringGetLength(tz_name);
+    CFIndex max_size = CFStringGetMaximumSizeForEncoding(length, encoding);
+    std::vector<char> buffer(max_size + 1);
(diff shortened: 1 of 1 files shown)


wada314:
Thank you so much!! I'll go to my office few hours later so I can check it at there. I'll reply again when I confirmed it's working.

src/time_zone_lookup.cc
```diff
@@ -126,8 +126,15 @@ time_zone local_time_zone() {
   _dupenv_s(&tz_env, nullptr, "TZ");
 #elif defined(__APPLE__)
   CFTimeZoneRef system_time_zone = CFTimeZoneCopyDefault();
-  CFStringRef tz_name = CFTimeZoneGetName(system_time_zone);
-  tz_env = strdup(CFStringGetCStringPtr(tz_name, CFStringGetSystemEncoding()));
+  if (CFStringRef tz_name = CFTimeZoneGetName(system_time_zone)) {
+    CFStringEncoding encoding = kCFStringEncodingUTF8;
+    CFIndex length = CFStringGetLength(tz_name);
+    CFIndex max_size = CFStringGetMaximumSizeForEncoding(length, encoding);
+    std::vector<char> buffer(max_size + 1);
```
devbww: `#include <vector>`

wada314:
I confirmed it's working in my Japanese mac. Thank you for the quick fix!

PR #102, opened 2019-04-02
Link to CoreFoundation in time_zone

JonathanDCohen:
This broke Abseil's cmake and Bazel builds, eg https://github.com/abseil/abseil-cpp/issues/291

This is required to fix the Bazel builds as well.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -15,16 +15,18 @@
 licenses(["notice"])  # Apache License
 
 config_setting(
-      name = "osx",
-      constraint_values = [
-          "@bazel_tools//platforms:osx"
-      ]
-  )config_setting(
-      name = "ios",
-      constraint_values = [
-          "@bazel_tools//platforms:ios",
-      ]
-  )
+    name = "osx",
+    constraint_values = [
+        "@bazel_tools//platforms:osx"
+    ],
+)
+
+config_setting(
+    name = "ios",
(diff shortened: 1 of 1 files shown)


BUILD
```diff
@@ -14,6 +14,20 @@
...
+    name = "osx",
+    constraint_values = [
+        "@bazel_tools//platforms:osx"
+    ],
+)
+
+config_setting(
+    name = "ios",
+    constraint_values = [
+        "@bazel_tools//platforms:ios",
```
devbww: (14:02:24) ERROR: /Volumes/BuildData/tmpfs/src/github/cctz/BUILD:24:1: no such target '@bazel_tools//platforms:ios': target 'ios' not declared in package 'platforms' (did you mean 'os'?) defined by /Volumes/BuildData/tmpfs/tmp/bazel/5f55c9a5875670d2927b9d5d4a6e1455/external/bazel_tools/platforms/BUILD and referenced by '//:ios'

devbww: Hey Jon: We're looking to cut a CCTZ release, and it would be nice to have this in.

devbww: We fixed the MacOS build to not use an old bazel.

JonathanDCohen: for posterity -- the issue was a bug in the presubmits accidentally using an old version of Bazel.

PR #105, opened 2019-04-29
A 5X performance improvement in next_weekday()/prev_weekday()

devbww:
Only do one get_weekday() call and one CivilDay addition per
next_weekday()/prev_weekday() call instead of one pair for
every day between the input and output.

Add benchmarks for get_weekday()/next_weekday()/prev_weekday().

### DIFF
--- a/include/cctz/civil_time_detail.h
+++ b/include/cctz/civil_time_detail.h
@@ -517,13 +517,43 @@ CONSTEXPR_F weekday get_weekday(const civil_day& cd) noexcept {
 ////////////////////////////////////////////////////////////////////////
 
 CONSTEXPR_F civil_day next_weekday(civil_day cd, weekday wd) noexcept {
-  do { cd += 1; } while (get_weekday(cd) != wd);
-  return cd;
+  CONSTEXPR_D weekday k_weekdays_forw[14] = {
+      weekday::monday,    weekday::tuesday,  weekday::wednesday,
+      weekday::thursday,  weekday::friday,   weekday::saturday,
+      weekday::sunday,    weekday::monday,   weekday::tuesday,
+      weekday::wednesday, weekday::thursday, weekday::friday,
+      weekday::saturday,  weekday::sunday,
+  };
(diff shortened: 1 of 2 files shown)


include/cctz/civil_time_detail.h
```diff
@@ -517,13 +517,43 @@ CONSTEXPR_F weekday get_weekday(const civil_day& cd) noexcept {
...
-  return cd;
+  CONSTEXPR_D weekday k_weekdays_forw[14] = {
+      weekday::monday,    weekday::tuesday,  weekday::wednesday,
+      weekday::thursday,  weekday::friday,   weekday::saturday,
+      weekday::sunday,    weekday::monday,   weekday::tuesday,
+      weekday::wednesday, weekday::thursday, weekday::friday,
+      weekday::saturday,  weekday::sunday,
+  };
+  weekday base = get_weekday(cd);
+  for (int i = 0;; ++i) {
```
devjgm: Would it help us in anyway way to use the fact that we know the integer value of the weekday enum since we are the ones implementing the library? That way, we could implement this with something like:

```cc
const auto n = static_cast<int>(get_weekday(cd));
const auto x = static_cast<int>(wd);
return cd + 1 + (x + 6 - n) % 7;
```

devbww: I benchmark the new `next/prev_weekday()` at ~14ns, which is about the same as `++cd`.  So, I don't think the difference between the two is discernible, and therefore would choose to avoid baking in that knowledge.

[Aside: You probably want an extra () pair in your example, to only do one CivilDay addition.]

devjgm: [Yes, to the extra parens.]

I wasn't trying to improve the performance any more, just seeing if the implementation could be simplified: 16 lines vs 3 lines. I don't see a problem w/ the library relying on its own implementation detail. But I don't feel strongly about this. Just wanted to mention this option. I'm happy with your PR if that's what you'd prefer.

PR #108, opened 2019-05-20
Simplify the handling of the tm_gmtoff and tm_zone extension fields

devbww:
The support for "libc:localtime/UTC" uses platform-specific mechanisms
to populate the time_zone::absolute_lookup::offset and abbr fields.
Here we simplify the access to those mechanisms a little, including
better adaptation to the `struct tm` extension fields.

### DIFF
--- a/src/time_zone_libc.cc
+++ b/src/time_zone_libc.cc
@@ -32,7 +32,7 @@ namespace {
 
 #if defined(_WIN32) || defined(_WIN64)
 // Uses the globals: '_timezone', '_dstbias' and '_tzname'.
-auto tm_gmtoff(const std::tm& tm) -> decltype(_timezone) {
+auto tm_gmtoff(const std::tm& tm) -> decltype(_timezone + _dstbias) {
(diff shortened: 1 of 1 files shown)


src/time_zone_libc.cc
```diff
@@ -31,57 +30,67 @@ namespace cctz {
...
+}
+auto tm_zone(const std::tm& tm) -> decltype(tzname[0]) {
+  const bool is_dst = tm.tm_isdst > 0;
+  return tzname[is_dst];
 }
 #elif defined(__native_client__) || defined(__myriad2__) || \
     defined(__EMSCRIPTEN__)
 // Uses the globals: 'timezone' and 'tzname'.
-OffsetAbbr get_offset_abbr(const std::tm& tm) {
+auto tm_gmtoff(const std::tm& tm) -> decltype(_timezone) {
```
devjgm: Do you think it's worth making the trailing return type `decltype(_timezone + 0)` to include the arithmetic with an int in the computed type?

devbww: Done.

PR #110, opened 2019-05-29
Fix -Wunused-member-function warning.

derekmauro:
XCode/Clang complains that the constructor to
AndroidZoneInfoSource is unused.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -664,6 +664,7 @@ std::unique_ptr<ZoneInfoSource> FileZoneInfoSource::Open(
   return std::unique_ptr<ZoneInfoSource>(new FileZoneInfoSource(fp, length));
 }
 
+#if defined(__ANDROID__)
 class AndroidZoneInfoSource : public FileZoneInfoSource {
  public:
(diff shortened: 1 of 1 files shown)


src/time_zone_info.cc
```diff
@@ -735,7 +736,9 @@ bool TimeZoneInfo::Load(const std::string& name) {
   auto zip = cctz_extension::zone_info_source_factory(
       name, [](const std::string& name) -> std::unique_ptr<ZoneInfoSource> {
         if (auto zip = FileZoneInfoSource::Open(name)) return zip;
+#if defined(__ANDROID__)
```
devbww: I'm thinking that the cure may be worse than the symptom.

When I first considered this some time back, I guarded the decl/defn of the (private) AndroidZoneInfoSource ctor instead, but now I'm thinking it might be better to simply remove the __ANDROID__ conditionals altogether.  There is nothing really Android specific about the code ... it is just the file names and the shape of the data that differ, and any platform could have appropriately shaped data in those locations.

What do you think?

derekmauro: SGTM. Unconditional compiling also has the advantage of preventing bit-rot, especially since getting an Android continuous build isn't easy.

PR #114, opened 2019-06-13
Make use of strptime POSIX-compliant

deeglaze:
Using #define before #include is considered bad C++ practice, so the
preprocessor definitions are moved to BUILD.

Reasoning from marmstrong@ on the Google C++ core libraries team: ""I [think] the approach of setting defines before #include is not good for C++.  The approach tends to break C++ modules (go/modules) and, less importantly, breaks clang-format's management of the include blocks in code.  It is also good to support moving code from .cc files to .h files (to support inline and template functions)."

Signed-off-by: Dionna Glaze <dionnaglaze@google.com>

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -28,6 +28,13 @@ config_setting(
     ],
 )
 
+config_setting(
+    name = "windows",
+    constraint_values = [
+        "@bazel_tools//platforms:windows",
+    ],
+)
+
 ### libraries
 
 cc_library(
@@ -65,6 +72,15 @@ cc_library(
         "include/cctz/time_zone.h",
         "include/cctz/zone_info_source.h",
     ],
+    copts = select({
+        "//:windows": [],
+        "//conditions:default": [
+            # Assume all non-windows platforms have strptime.
+            "-DHAS_STRPTIME=1",
+            # Include strptime from time.h.
+            "-D_XOPEN_SOURCE=700",
+        ],
+    }),
     includes = ["include"],
     linkopts = select({
         "//:osx": [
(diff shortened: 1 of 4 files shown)


BUILD
```diff
@@ -28,6 +28,13 @@ config_setting(
     ],
 )
 
+config_setting(
+    name = "windows",
+    constraint_values = [
+        "@bazel_tools//platforms:windows",
```
devbww: Does this encompass MinGW?

deeglaze: Yes it does https://github.com/bazelbuild/bazel/blob/master/tools/cpp/BUILD.windows.tpl#L130

deeglaze:
Conditional exports in C based on preprocessor definitions are the antithesis to the hermetic build goals of Bazel. Every time a header has to be parsed in a different way given different definitions, it has to be considered a different library's header.

Since strptime isn't a C++ standard function, _XOPEN_SOURCE is well-documented, and #define before #include is commonplace for POSIX-standard includes, I'm not dead-set on moving these defines to BUILD/Makefile/CMakeLists.txt if your preference is strongly towards the existing method.

devbww:
> Conditional exports in C based on preprocessor definitions are the antithesis to the hermetic build goals of Bazel. Every time a header has to be parsed in a different way given different definitions, it has to be considered a different library's header.

But there are no headers being parsed in different ways here.  It is just a single implementation file choosing whether to include a header or not.

We could make `src/time_zone_format.cc` unconditionally include `<iomanip>` and `<sstream>` if you like.

deeglaze:
> But there are no headers being parsed in different ways here. It is just a single implementation file choosing whether to include a header or not.

I'm just talking about how time.h gets interpreted differently based on the definedness of _XOPEN_SOURCE, because it will declare strptime only if it's defined.

deeglaze:
I created https://github.com/google/cctz/pull/115 for an alternative for you to select by your preference.

deeglaze:
The alternative looks preferred. Closing.

PR #115, opened 2019-06-18
Define _XOPEN_SOURCE to import strptime

deeglaze:
POSIX standards require _XOPEN_SOURCE to be defined with any value to
import strptime from time.h. Since the C++ standard doesn't specify
strptime, ctime does not have to be implemented with #include <time.h>.
Therefore, this change adds the include directly.

This PR is an alternative to https://github.com/google/cctz/pull/114

### DIFF
--- a/src/time_zone_format.cc
+++ b/src/time_zone_format.cc
@@ -18,7 +18,17 @@
 # endif
 #endif
 
+#if defined(HAS_STRPTIME) && HAS_STRPTIME
+# if !defined(_XOPEN_SOURCE)
+#  define _XOPEN_SOURCE  // Definedness suffices for strptime.
+# endif
+#endif
+
 #include "cctz/time_zone.h"
+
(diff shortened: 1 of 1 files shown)


src/time_zone_format.cc
```diff
@@ -18,6 +18,15 @@
 # endif
 #endif
 
+#if HAS_STRPTIME
+# if !defined(_XOPEN_SOURCE)
+#  define _XOPEN_SOURCE  // Definedness suffices for strptime.
+# endif
+// Include time.h directly since, by C++ standards, ctime doesn't have to
```
devbww: Should we move this `#include` part down to the `HAS_STRPTIME` section in the includes below?

deeglaze: Done. The defines have to stay above the first #include though, since it transitively #include's chrono, which can include ctime, which can include time.h :|

devbww:
Thanks.

PR #123, opened 2019-08-09
Ensure we don't call destructor for global static time_zone_mutex

bhamiltoncx:
time_zone_impl.cc contains a global static std::mutex to protect the mutable time_zone_map.

On platforms where std::mutex's destructor is non-trivial (see https://bugs.llvm.org/show_bug.cgi?id=27658), the global mutex is at risk of the static deinitialization order fiasco.

This CL fixes the issue by using `new` to move the file-scoped std::mutex onto the heap, intentionally leaking it to avoid the static deinitialization order fiasco.

Tested:
  % bazel test :all

### DIFF
--- a/src/time_zone_impl.cc
+++ b/src/time_zone_impl.cc
@@ -31,7 +31,9 @@ using TimeZoneImplByName =
 TimeZoneImplByName* time_zone_map = nullptr;
 
 // Mutual exclusion for time_zone_map.
-std::mutex time_zone_mutex;
+// This mutex is intentionally "leaked" to avoid the static deinitialization
+// order fiasco (std::mutex's destructor is not trivial on many platforms).
+std::mutex* time_zone_mutex = new std::mutex;
 
 }  // namespace
 
@@ -52,7 +54,7 @@ bool time_zone::Impl::LoadTimeZone(const std::string& name, time_zone* tz) {
   // Then check, under a shared lock, whether the time zone has already
   // been loaded. This is the common path. TODO: Move to shared_mutex.
   {
-    std::lock_guard<std::mutex> lock(time_zone_mutex);
+    std::lock_guard<std::mutex> lock(*time_zone_mutex);
     if (time_zone_map != nullptr) {
       TimeZoneImplByName::const_iterator itr = time_zone_map->find(name);
       if (itr != time_zone_map->end()) {
@@ -63,7 +65,7 @@ bool time_zone::Impl::LoadTimeZone(const std::string& name, time_zone* tz) {
   }
 
   // Now check again, under an exclusive lock.
-  std::lock_guard<std::mutex> lock(time_zone_mutex);
+  std::lock_guard<std::mutex> lock(*time_zone_mutex);
   if (time_zone_map == nullptr) time_zone_map = new TimeZoneImplByName;
   const Impl*& impl = (*time_zone_map)[name];
   if (impl == nullptr) {
@@ -82,7 +84,7 @@ bool time_zone::Impl::LoadTimeZone(const std::string& name, time_zone* tz) {
 }
 
 void time_zone::Impl::ClearTimeZoneMapTestOnly() {
-  std::lock_guard<std::mutex> lock(time_zone_mutex);
+  std::lock_guard<std::mutex> lock(*time_zone_mutex);
   if (time_zone_map != nullptr) {
     // Existing time_zone::Impl* entries are in the wild, so we simply
     // leak them.  Future requests will result in reloading the data.


bhamiltoncx:
There is a tradeoff here: we're avoiding a startup time hit (paging the segments of the binary from disk into memory) and adding a branch-predicted atomic fetch to check the state of the C++ static "constructed only once" bit.

bhamiltoncx:
Here's a godbolt (clang trunk -O3) showing a comparison:

https://godbolt.org/z/ZJWteq

I'm not sure how hot this code is, but it's good to understand the tradeoffs here.

src/time_zone_impl.cc
```diff
@@ -30,8 +30,14 @@ using TimeZoneImplByName =
     std::unordered_map<std::string, const time_zone::Impl*>;
 TimeZoneImplByName* time_zone_map = nullptr;
 
-// Mutual exclusion for time_zone_map.
-std::mutex time_zone_mutex;
+std::mutex& TimeZoneMutex() {
+  // Mutual exclusion for time_zone_map.
```
devbww: Let's move this comment back outside.

bhamiltoncx: Done.

devbww:
> There is a tradeoff here: we're avoiding a startup time hit (paging the segments of the binary from disk into memory) and adding a branch-predicted atomic fetch to check the state of the C++ static "constructed only once" bit.

Yes.

I'm a little surprised that the startup hit is discernible, but I guess it depends on when you load your first timezone, what else shares a page with that mutex, and other subtleties of your platform.  Let me just confirm with you that the change really makes a difference for issue 1.

On the "constructed-bit" side I'm a little concerned (particularly so on non-x86).  It seems like we're doubling the synchronization cost.  This is just in `LoadTimeZone()` though, and if that is a hot spot then the application probably has bigger problems.

So, I'm all for fixing issue 2, but am hoping that ...
```
std::mutex* time_zone_mutex = new std::mutex;
```
might be sufficient (that is, that issue 1 is ignorable).

Thanks.

bhamiltoncx:
> I'm a little surprised that the startup hit is discernible, but I guess it depends on when you load your first timezone, what else shares a page with that mutex, and other subtleties of your platform. Let me just confirm with you that the change really makes a difference for issue 1.

Ah, let me provide a bit more context: this is for code which does not directly use Abseil *or* cctz, but instead pulls it in via dependencies of dependencies of dependencies.

The project I'm working on doesn't actually load timezones or even touch any of this code, but since absl and cctz are such widely-used dependencies, every downstream client has to pay the startup cost synchronously before `main()` can even begin.

In reality, the dependent library code which does touch cctz is invoked on a background thread, so we're trying to move the cost of paging in the segments of the binary which include absl to that thread.

We did measure this in an A/B test (20 runs each on 3 different devices with and without this code linked in) and found it's roughly 1 ms of additional pre-main startup time (opt build) for each static C++ initializer our dependencies invoke.

Of course, it's not the initializer actually taking 1 ms of CPU time, but the cost is dominated by paging in from disk the cctz and Abseil code and its dependencies.

> On the "constructed-bit" side I'm a little concerned (particularly so on non-x86). It seems like we're doubling the synchronization cost. This is just in `LoadTimeZone()` though, and if that is a hot spot then the application probably has bigger problems.

I believe the cost is going to be dominated by the function call to `pthread_mutex_lock`. The extra code in https://godbolt.org/z/ZJWteq boils down to:

```
       mov     al, byte ptr [rip + guard variable for (anonymous namespace)::BarMutex()::bar_mutex]
        test    al, al
        je      .LBB0_1
```

which I'm fairly certain will be drowned out by the immediately-following code to invoke `pthread_mutex_lock()`:

```
        mov     rbx, qword ptr [rip + (anonymous namespace)::BarMutex()::bar_mutex]
        mov     rdi, rbx
        call    pthread_mutex_lock
```

devbww:
> the dependent library code which does touch cctz is invoked on a background thread, so we're trying to move the cost of paging in the segments of the binary which include absl to that thread.

It sounds like you want some form of dynamic linking, where you don't run the library's global constructors until the library is needed and loaded.

> I believe the cost is going to be dominated by the function call to pthread_mutex_lock. The extra code in https://godbolt.org/z/ZJWteq boils down to ...

On non-x86 platforms this becomes much more complicated (introducing memory fences, for example), essentially creating a special-purpose, synchronized flag.

But, as I mentioned, this is just in `LoadTimeZone()`, so perhaps I'm being overly paranoid.

bhamiltoncx:
Discussed offline with @devbww . Updated to only solve the static deinitialization order fiasco for now so we can focus on correctness before performance.

I'd still like to follow up to resolve the pre-main static initializer, but we can do that separately.

PR #124, opened 2019-08-13
Fix SIOF issue by moving time_zone_mutex into function

bhamiltoncx:
Originally, https://github.com/google/cctz/pull/123 moved time_zone_mutex
into a function to solve both a SDOF issue as well as a performance issue.

The PR was restructured to only fix the SDOF issue, but that
introduced a SIOF issue where time_zone_mutex was moved to a new'd
pointer on the heap instead of a link-time initialized std::mutex.

To resolve the SIOF issue, this PR restores the original behavior of
moving time_zone_mutex into a function. This does introduce an
additional atomic load (or, on MSVC, a TLS access) to check if
the mutex is initialized, but this is not expected to be a hot
code-path since it can potentially hit the disk to load a time zone
definition.

Tested:
  % bazel test :all

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/124)

### DIFF
--- a/src/time_zone_impl.cc
+++ b/src/time_zone_impl.cc
@@ -31,9 +31,12 @@ using TimeZoneImplByName =
 TimeZoneImplByName* time_zone_map = nullptr;
 
 // Mutual exclusion for time_zone_map.
-// This mutex is intentionally "leaked" to avoid the static deinitialization
-// order fiasco (std::mutex's destructor is not trivial on many platforms).
-std::mutex* time_zone_mutex = new std::mutex;
+std::mutex& TimeZoneMutex() {
+  // This mutex is intentionally "leaked" to avoid the static deinitialization
+  // order fiasco (std::mutex's destructor is not trivial on many platforms).
+  static std::mutex* time_zone_mutex = new std::mutex;
+  return *time_zone_mutex;
+}
 
 }  // namespace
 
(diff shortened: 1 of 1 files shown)


src/time_zone_impl.cc
```diff
@@ -31,9 +31,12 @@ using TimeZoneImplByName =
 TimeZoneImplByName* time_zone_map = nullptr;
 
 // Mutual exclusion for time_zone_map.
-// This mutex is intentionally "leaked" to avoid the static deinitialization
-// order fiasco (std::mutex's destructor is not trivial on many platforms).
-std::mutex* time_zone_mutex = new std::mutex;
+std::mutex& TimeZoneMutex() {
+  // This mutex is intentionally "leaked" to avoid the static deinitialization
+  // order fiasco (std::mutex's destructor is not trivial on many platforms).
+  static std::mutex time_zone_mutex;
```
devbww: Won't this still have its destructor run?  That is, there is no "leak" here.

bhamiltoncx: Of course, my mistake. Returned to making this a pointer (as it was in the original PR) to avoid SDOF.

src/time_zone_impl.cc
```diff
@@ -31,9 +31,12 @@ using TimeZoneImplByName =
 TimeZoneImplByName* time_zone_map = nullptr;
 
 // Mutual exclusion for time_zone_map.
-// This mutex is intentionally "leaked" to avoid the static deinitialization
-// order fiasco (std::mutex's destructor is not trivial on many platforms).
-std::mutex* time_zone_mutex = new std::mutex;
+std::mutex& TimeZoneMutex() {
+  // This mutex is intentionally "leaked" to avoid the static deinitialization
+  // order fiasco (std::mutex's destructor is not trivial on many platforms).
+  static std::mutex *time_zone_mutex = new std::mutex;
```
devbww: Shift the `*` from the variable name to the type name.

bhamiltoncx: Whoops! Fixed, thanks.

PR #125, opened 2019-08-15
Load @rules_cc for compatibility with Bazel 1.0

derekmauro:
Starting with Bazel 1.0 (September 2019), C++ rules will need to be
loaded from the @rules_cc repository. This change adds the required
loads for that.

Pull request for Abseil is
https://github.com/abseil/abseil-cpp/pull/351

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/125)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -12,6 +12,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
+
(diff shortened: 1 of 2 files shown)


derekmauro:
When I tried to import the Abseil pull request, I found that the version of Bazel on Kokoro Windows and macOS was too old to support this.

This is going to be interesting.

derekmauro:
After updating the test VMs, the tests are now passing.

PR #139, opened 2020-04-14
Remove dependency on CoreFoundation library

alexey-milovidov:
Dependency on OS library is unwanted. It works fine without it.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/139)


alexey-milovidov:
Sorry, I've opened PR in a wrong repository.

PR #142, opened 2020-05-09
Migrate parsing to not rely on null characters

fowles:
Migrate cctz to not rely on null termination in most of its internal parsing details.  Existing public api quickly shims onto the existing work without much change.

Benchmarks are neutral.

| Benchmark                                 |         Time           | CPU | 
| ----------------------------------------|------------------------|-------- |            
| BM_Difference_Days                            | -0.0001 |        -0.0002 |
| BM_Step_Days       |                             +0.0006 |        +0.0005  |
| BM_GetWeekday     |                              +0.0000        | -0.0001 |
| BM_NextWeekday    |                              +0.0002       |  +0.0001 | 
| BM_PrevWeekday     |                             +0.0008      |   +0.0006 |
| BM_Zone_LoadUTCTimeZoneFirst   |                 +0.0029     |    +0.0028 |
| BM_Zone_LoadUTCTimeZoneLast    |                 +0.0007   |      +0.0006|
| BM_Zone_LoadTimeZoneFirst            |           +0.0026       |  +0.0023     |
| BM_Zone_LoadTimeZoneCached        |              +0.0011      |   +0.0009 |
| BM_Zone_LoadLocalTimeZoneCached |                +0.0049 |        +0.0049 |
| BM_Zone_LoadAllTimeZonesFirst          |         +0.0018      |   +0.0019|
| BM_Zone_LoadAllTimeZonesCached      |            +0.0116 |        +0.0116|          
| BM_Zone_TimeZoneEqualityImplicit         |       -0.0000  |       -0.0000      |      
| BM_Zone_TimeZoneEqualityExplicit          |      -0.0046 |        -0.0047       |     
| BM_Zone_UTCTimeZone  |                           -0.0028 |        -0.0028           |  
| BM_Time_ToCivil_CCTZ     |                       -0.0059 |        -0.0060            |
| BM_Time_ToCivil_Libc         |                   -0.0007   |      -0.0007            |
| BM_Time_ToCivilUTC_CCTZ |                        +0.0122 |        +0.0121   |
| BM_Time_ToCivilUTC_Libc     |                    -0.0015   |      -0.0016        |
| BM_Time_FromCivil_CCTZ     |                     -0.0021 |        -0.0021      |
| BM_Time_FromCivil_Libc         |                 -0.0005 |        -0.0005         |
| BM_Time_FromCivilUTC_CCTZ |                       +0.0009  |       +0.0009|
| BM_Time_FromCivilDay0_CCTZ |                     +0.0074  |       +0.0074|
| BM_Time_FromCivilDay0_Libc     |                 -0.0038     |    -0.0038     |
| BM_Format_FormatTime/0            |              +0.0044     |    +0.0044     |
| BM_Format_FormatTime/1             |             +0.0011    |     +0.0010     |
| BM_Format_FormatTime/2              |            +0.0028   |      +0.0027     |
| BM_Format_FormatTime/3               |           +0.0042  |       +0.0039     |
| BM_Format_FormatTime/4                |          +0.0030       |  +0.0030     |
| BM_Format_FormatTime/5                 |         -0.0069       |  -0.0068       |
| BM_Format_ParseTime/0                    |       -0.0273       |  -0.0270        |
| BM_Format_ParseTime/1                     |      +0.0015     |    +0.0015      |
| BM_Format_ParseTime/2                      |     -0.0006     |    -0.0007        |
| BM_Format_ParseTime/3                       |    +0.0009   |      +0.0009      |
| BM_Format_ParseTime/4                        |   -0.0007   |      -0.0007       |
| BM_Format_ParseTime/5  |                         -0.0002  |       -0.0002        |

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/142)

### DIFF
--- a/include/cctz/string_detail.h
+++ b/include/cctz/string_detail.h
@@ -25,6 +25,8 @@ namespace detail {
 struct char_range {
   char_range(const char* s)
       : begin(s), end(s + strlen(s)) {}
+  char_range(const char* begin, const char* end)
+      : begin(begin), end(end) {}
   char_range(const std::string& s)
       : begin(s.data()), end(s.data() + s.size()) {}
   char_range(const char_range&) = default;
--- a/include/cctz/string_detail.h
+++ b/include/cctz/string_detail.h
@@ -31,18 +31,24 @@ struct char_range {
 
   friend bool operator==(char_range lhs, char_range rhs) {
     if (lhs.begin == rhs.begin && lhs.end == rhs.end) return true;
-    if (lhs.end - lhs.begin != rhs.end - rhs.begin) return false;
-    return memcmp(lhs.begin, rhs.begin, lhs.end - lhs.begin) == 0;
+    if (lhs.size() != rhs.size()) return false;
+    return memcmp(lhs.begin, rhs.begin, lhs.size()) == 0;
   }
 
   friend bool operator!=(char_range lhs, char_range rhs) {
     return !(lhs == rhs);
   }
 
-  bool starts_with(char_range s) {
-    if (end - begin < s.end - s.begin) return false;
-    return memcmp(begin, s.begin, s.end - s.begin) == 0;
+  bool starts_with(char_range s) const {
+    if (size() < s.size()) return false;
+    return memcmp(begin, s.begin, s.size()) == 0;
   }
+  bool starts_with(char c) const {
+    if (begin == end) return false;
+    return *begin == c;
+  }
+
+  size_t size() const { return end - begin; }
 
   const char* begin;
   const char* end;
--- a/include/cctz/string_detail.h
+++ b/include/cctz/string_detail.h
@@ -43,9 +43,18 @@ struct char_range {
     if (size() < s.size()) return false;
     return memcmp(begin, s.begin, s.size()) == 0;
   }
+
   bool starts_with(char c) const {
     if (begin == end) return false;
     return *begin == c;
+  }
+
+  bool consume_prefix(char c) {
+    if (starts_with(c)) {
+      ++begin;
+      return true;
+    }
+    return false;
   }
 
   size_t size() const { return end - begin; }
--- a/include/cctz/string_detail.h
+++ b/include/cctz/string_detail.h
@@ -39,6 +39,11 @@ struct char_range {
     return !(lhs == rhs);
   }
 
+  bool starts_with(char_range s) {
+    if (end - begin < s.end - s.begin) return false;
+    return memcmp(begin, s.begin, s.end - s.begin) == 0;
+  }
+
   const char* begin;
   const char* end;
 };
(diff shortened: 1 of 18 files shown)


include/cctz/string_detail.h
```diff
@@ -0,0 +1,90 @@
...
+#define CCTZ_CIVIL_STRING_DETAIL_H_
+
+#include <cctype>
+#include <cstring>
+#include <string>
+
+namespace cctz {
+namespace detail {
+
+struct char_range {
```
devbww, line 25: It is strange for `string_detail.h` to declare `char_range`.  Why not `char_range.h`?

But that said, why not implement `cctz::string_view` with whatever subset of the `std::string_view` interface you need, so that we can simply move to the latter if/when possible?

fowles: frankly, I sort of arrived at this incrementally, so it wasn't deeply thought out.  I had intended to make something with a much smaller interface than string_view so it didn't have as much baggage to go with it.

Would you like me to change to something like cctz::string_view?

src/time_zone_impl.cc
```diff
@@ -45,7 +45,8 @@ time_zone time_zone::Impl::UTC() {
   return time_zone(UTCImpl());
 }
 
-bool time_zone::Impl::LoadTimeZone(const std::string& name, time_zone* tz) {
+bool time_zone::Impl::LoadTimeZone(detail::char_range name_range, time_zone* tz) {
```
devbww, line 48: Given that these interfaces immediately turn around and create a `std::string` (at least for the foreseeable future), there doesn't seem to be much point in modifying the `LoadTimeZone()` path.

Indeed, allow me to restate an earlier question:  Is there evidence that we're spending significant amounts of time converting non-`std::string`s to `std::string`s such that we need to push that non-conversion through.  (Naively, I'd be surprised if that conversion was the dominating cost of formatting/parsing.)

include/cctz/string_detail.h
devbww: I think so, yes.  There is no need to implement the entire interface though ... just the subset you need to get the job at hand done (i.e., nothing more than you already have).

devbww: Greg mentioned that he wasn't sure what I had said here, so allow me to restate more clearly.

Add `include/cctz/string_view.h`, which implements `cctz::string_view` with a strict subset of the `std::string_view` interface.  Then use `cctz::string_view` in whatever interfaces you chose, safe in the knowledge that we'll be able to switch to `std::string_view` in the future.

I'm even happy to do that if you like.

fowles: if you are happy to take it over, that would be great.  I tried to do this for a few hours and discovered that the existing parser traffics in `const char*` very very hard, so migrating it to something that does not expose `const char*` public data members is quite awkward

devbww:
Closing until this repo can at least start using `std::string_view` (C++17).

PR #144, opened 2020-05-22
support parsing 'T' and 'Z' case insensitively

### LINKED ISSUE #143: Feature Request: Support parsing 'T' and 'Z' case insensitively
RFC 3339 [[link](https://tools.ietf.org/html/rfc3339#section-5.6)] allows the `T` and `Z` characters to be either upper or lower case. That means, both of the following are valid RFC 3339 strings:

```
      1985-04-12T23:20:50.52Z  // OK
      1985-04-12t23:20:50.52z  // OK
``` 

Today, there's no way to specify a format string that will parse either of those. We already have `%Ez` that can parse the UTC offset, and it also consumes the `Z`. Perhaps that could also be taught how to handle a lower-case `z`.

Maybe we similarly need a `%ET` extended format specifier that consumes either `T` or `t`. Just a thought.

Today, an alternative is for callers to try parsing the string 4x with various forms of upper/lower case T/Z.

devbww:
%EZ now accepts 'z' in addition to 'Z' as a synonym for +00:00.

%ET is introduced, producing 'T' on output, and accepting 'T'
or 't' on input.  This is for the RFC3339 date-time separator.

Fixes #143.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/144)

### DIFF
--- a/include/cctz/time_zone.h
+++ b/include/cctz/time_zone.h
@@ -287,6 +287,7 @@ bool parse(const std::string&, const std::string&, const time_zone&,
 //   - %E#f - Fractional seconds with # digits of precision
 //   - %E*f - Fractional seconds with full precision (a literal '*')
 //   - %E4Y - Four-character years (-999 ... -001, 0000, 0001 ... 9999)
+//   - %ET  - The RFC3339 "date-time" separator "T"
 //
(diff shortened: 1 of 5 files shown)


devbww:
No worries ... I already had it drafted.

PR #162, opened 2020-08-22
Use libc:localtime for local_time_zone on Emscripten instead of hitting the filesystem.

suertreus:
Emscripten doesn't necessarily link in filesystem support nor have zoneinfo files available, but its libc does know the local time zone.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/162)

### DIFF
--- a/src/time_zone_lookup.cc
+++ b/src/time_zone_lookup.cc
@@ -139,6 +139,11 @@ time_zone local_time_zone() {
   }
   CFRelease(tz_default);
 #endif
+#if defined(__EMSCRIPTEN__)
+  // Emscripten doesn't necessarily link in filesystem support nor have zoneinfo
+  // files available, so use libc (which does work) instead.
+  zone = "libc:localtime";
+#endif
 
   // Allow ${TZ} to override to default zone.
   char* tz_env = nullptr;


suertreus:
> My first question would be, how does libc work? If it supports "localtime" for any `TZ=<zone>`, then it must have some source of transition data.

I'm a tourist in both codebases, but it looks to me like (in my configuration at least) it asks the JS runtime (e.g. the browser or nodejs):
https://github.com/emscripten-core/emscripten/blob/master/src/library.js#L1163

It doesn't look like this codepath respects `$TZ`, nor does it look like the required get-a-timezone-by-name JS API exists:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

For the same reason, I don't think we can implement a `ZoneInfoSource` in cctz that's able to handle timezones other than local and UTC.  If you want, I can take a swing at making one for the Emscripten platform that's limited in that way, but it's pretty much just going to be a copy of the source that implements `libc:localtime`.

I can also work around this on my end by looking up `libc:localtime` explicitly on `__EMSCRIPTEN__`, but that's obviously not going to get you any closer to deleting `libc:localtime`.

kripken:
> It doesn't look like this codepath respects $TZ (my tests confirm this)

(Emscripten dev here) Yes, I think that's right, our JS code just reads the browser's state. I'm not very familiar with timezones in general, but it might be possible to look at TZ and return valid values for that. That might increase code size, though - right now we can avoid shipping most of libc's time code thanks to relying on the JS environment. But if it's important we can open an issue on emscripten and discuss details.

suertreus:
In order to make sense of a TZ name like Asia/Tokyo, the program needs a zoneinfo database, plus code to interpret it.  Linking in the relevant bits of musl won't help unless we also bundle zoneinfo data or have some way to get them from the system.

tzdb is updated multiple times a year; I infer that any bundled data would become outdated pretty quickly.

devbww:
Thanks for the references and comments, Andy and Alon.  I'll try to step through them with the goal of solving your problem, but again my goal is to be able to remove `libc:localtime`.  CCTZ is supposed to be an alternative to libc, not a wrapper around it.  And I'm pretty sure we wouldn't invent `libc:localtime` to solve your problem if it hadn't already existed.

I can't say I'm particularly enthralled with the `library.js:localtime_r` implementation.  Trying to glean offset information from January 1st and June 1st, for example, is guesswork at best.  It is also still unclear where `Date` gets its localtime transitions/offsets from, or how it even knows what the local timezone is (i.e., I don't understand the "browser state"/"JS environment").  If we understood that, it may prove sufficient to craft an Emscripten `ZoneInfoSource`.

But back to the original Google bug report for a second.  It seems to be the result of a `absl::LocalTimeZone().At()` call in the logging code.  One simple alternative you might consider is using `absl::UTCTimeZone().At()` instead for Emscripten.  `UTCTimeZone()` won't try to read anything (so you shouldn't get that "should not be possible to operate on streams" error), and it may even be a feature for log entries to be stamped with the UTC time-of-day rather than local.  [Aside: I'm of the opinion we should use `UTCTimeZone()` unconditionally for logging, but I haven't won that argument yet. :-(]

Otherwise, it sounds like space might be a concern for embedding all the zoneinfo data in your binary.  At the moment it looks like that would amount to ~670kB uncompressed.  There is a new, "slim" format that shrinks that to ~326kB.  But we could reduce that even further, to ~80kB, if each zone only encoded its current POSIX rule (i.e., if the current DST rule was assumed to apply throughout history, which would certainly be fine if we're only talking about logging).  If any of those numbers are acceptable, embedding zoneinfo would be another alternative.

Anyway, there are three possibilities to think about.

> tzdb is updated multiple times a year; I infer that any bundled data would become outdated pretty quickly.

Yes, but if we're only worried about one particular zone (the "local" one), then it wouldn't be outdated any more often than whatever baked-in mechanism `Date` is using to determine the local rule.

devbww:
Another possibility you might consider for Emscripten CCTZ support is to make `FileZoneInfoSource::Open()` always fail, without trying to open anything.  Then all zones will fallback to UTC.

kripken:
Even an 80kB in binary size would be too much in Emscripten, at least by default. If there's no other way we could add an option for this. But even an option feels wrong to me personally - web pages don't normally need that extra information, or there would be a Web API for it? But maybe I don't understand the use cases for this - I don't actually know that much about time zones.

suertreus:
> It is also still unclear where Date gets its localtime transitions/offsets from, or how it even knows what the local timezone is (i.e., I don't understand the "browser state"/"JS environment").

I haven't looked this deep, but I assume these are wired into the browser, which I assume calls libc.

>> tzdb is updated multiple times a year; I infer that any bundled data would become outdated pretty quickly.
>
> Yes, but if we're only worried about one particular zone (the "local" one), then it wouldn't be outdated any more often than whatever baked-in mechanism Date is using to determine the local rule.

If that's indeed libc, it's reading zoneinfo files that are kept up to date by the OS.  Even if the OS is not updating them, at least it's not our fault if the data are stale and the log timestamp is wrong.

If we bundle zoneinfo data into the binary, we're stuck with them for however long the binary is used by users, which is ~unbounded for Abseil users in general.

**Punchline:**
For now, I think it best to close this PR and have my code use `localtime_r` on Emscripten.  I'll note in a comment that this platform typically does not have access to zoneinfo data and so should avoid using cctz for timezones.

Aspirational thoughts:
- This looks like it could be a long-term solution, one day:
https://tc39.es/proposal-temporal/docs/timezone.html
- I agree that UTC would be better.  I'll take one last look at whether that's possible to change, but I am... not optimistic.

devbww:
> I agree that UTC would be better. I'll take one last look at whether that's possible to change, but I am... not optimistic.

I'm only suggesting `UTCTimeZone()` for Emscripten.  Everyone else can continue to use `LocalTimeZone()` (and maybe one day they'll see the light).

suertreus:
I think I prefer an `#ifdef` over having one platform behave differently, but I'll do a straw-poll on chat in case that preference is not widely shared.

Thanks for considering this change and explaining your position to me.

devbww:
> I think I prefer an `#ifdef` over having one platform behave differently

I'm not sure I understand the distinction, but an `#ifdef` in the logging code is indeed what I am envisaging.

suertreus:
Ah, sorry, I meant an `#ifdef` with a call to `localtime_r` inside.  I can construct an `absl::CivilSecond` directly from the resulting `struct tm` and then rejoin the other platforms at the cool kids' table.

devbww:
OK.  Sounds good.  Thanks.

PR #164, opened 2020-08-24
Adds a missing linker flag (-lm) to allow building cctz on FreeBSD.

### LINKED ISSUE #163: Linker errors on FreeBSD (-lm missing)
```shell
❯ bazel test ...
WARNING: /usr/home/yesudeep/.cache/bazel/_bazel_yesudeep/12db47f080f7e9f258fb2857efef7faf/external/remote_java_tools_linux/BUILD:671:11: in hdrs attribute of cc_library rule @remote_java_tools_linux//:combiners: Artifact 'external/remote_java_tools_linux/java_tools/src/tools/singlejar/zip_headers.h' is duplicated (through '@remote_java_tools_linux//:transient_bytes' and '@remote_java_tools_linux//:zip_headers'). Since this rule was created by the macro 'cc_library', the error might have been caused by the macro implementation
INFO: Analyzed 17 targets (1 packages loaded, 651 targets configured).
INFO: Found 13 targets and 4 test targets...
ERROR: /usr/home/yesudeep/code/yesudeep/cctz/BUILD:227:10: Linking of rule '//:example2' failed (Exit 1) clang failed: error executing command /usr/bin/clang -o bazel-out/freebsd-fastbuild/bin/example2 bazel-out/freebsd-fastbuild/bin/_objs/example2/example2.pic.o bazel-out/freebsd-fastbuild/bin/libtime_zone.a ... (remaining 5 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
ld: error: undefined symbol: ceilf
>>> referenced by time_zone_impl.cc
>>>               time_zone_impl.pic.o:(ceil(float)) in archive bazel-out/freebsd-fastbuild/bin/libtime_zone.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)
INFO: Elapsed time: 0.513s, Critical Path: 0.08s
INFO: 1 process: 1 processwrapper-sandbox.
FAILED: Build did NOT complete successfully
//:civil_time_test                                              (cached) PASSED in 0.2s
//:cctz_benchmark                                                     NO STATUS
//:time_zone_format_test                                              NO STATUS
//:time_zone_lookup_test                                              NO STATUS

FAILED: Build did NOT complete successfully
```

yesudeep:
Fixes #163 when building cctz on FreeBSD.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/164)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -24,6 +24,13 @@ config_setting(
 )
 
 config_setting(
+    name = "openbsd",
+    constraint_values = [
+        "@bazel_tools//platforms:openbsd",
+    ],
+)
+
+config_setting(
     name = "osx",
     constraint_values = [
         "@bazel_tools//platforms:osx",
@@ -49,6 +56,10 @@ cc_library(
     textual_hdrs = ["include/cctz/civil_time_detail.h"],
     visibility = ["//visibility:public"],
 )
+
+BSD_LINKOPTS = [
+    "-lm",
+]
 
 cc_library(
     name = "time_zone",
@@ -82,9 +93,8 @@ cc_library(
         "//:ios": [
             "-framework Foundation",
         ],
-        "//:freebsd": [
-            "-lm",
-        ],
+        "//:freebsd": BSD_LINKOPTS,
+        "//:openbsd": BSD_LINKOPTS,
         "//conditions:default": [],
     }),
     visibility = ["//visibility:public"],
--- a/BUILD
+++ b/BUILD
@@ -17,6 +17,13 @@ load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 licenses(["notice"])  # Apache License
 
 config_setting(
+    name = "freebsd",
+    constraint_values = [
+        "@bazel_tools//platforms:freebsd",
+    ],
+)
+
+config_setting(
     name = "osx",
     constraint_values = [
         "@bazel_tools//platforms:osx",
@@ -74,6 +81,9 @@ cc_library(
         ],
         "//:ios": [
             "-framework Foundation",
+        ],
+        "//:freebsd": [
+            "-lm",
         ],
         "//conditions:default": [],
     }),


BUILD
```diff
@@ -75,6 +82,9 @@ cc_library(
         "//:ios": [
             "-framework Foundation",
         ],
+        "//:freebsd": [
+            "-lm",
+        ],
```
devbww, line 87: I don't see why this is required.  CCTZ is, after all, a C++ library, and it doesn't include `math.h`.

```
ld: error: undefined symbol: ceilf
>>> referenced by time_zone_impl.cc
>>>               time_zone_impl.pic.o:(ceil(float)) in archive bazel-out/freebsd-fastbuild/bin/libtime_zone.a
```

Whoever generated the call to `ceilf()` should be the one to add the dependency.  It is not made explicitly from `time_zone_impl`.

yesudeep: I do notice that building using `cmake` on both FreeBSD and Linux produces output that shows a dependency on `libm`. (`time_tool` depends on `:time_zone`):

Linux:
```shell
❯ ldd time_tool 
	linux-vdso.so.1 (0x00007ffefcc61000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007ff5b32c8000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ff5b32ad000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff5b30bb000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff5b2f6c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007ff5b3529000)
```

FreeBSD:
```
❯ ldd time_tool
time_tool:
        libc++.so.1 => /usr/lib/libc++.so.1 (0x800287000)
        libcxxrt.so.1 => /lib/libcxxrt.so.1 (0x800354000)
        libm.so.5 => /lib/libm.so.5 (0x800378000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x8003aa000)
        libc.so.7 => /lib/libc.so.7 (0x8003c1000)
```

And on FreeBSD/OpenBSD, for several projects, the `-lm` flag to the linker is usually explicit even if it is C++
source code. Here's an example from the [bazel source tree](https://github.com/bazelbuild/bazel/blob/ebf6a46e25d4be3bf0bbd77cb21dbff0276d5d04/src/tools/singlejar/BUILD#L76). I'm not sure whether CMake special cases this for FreeBSD/OpenBSD, however. 

Where do you reckon the flag would be more appropriately placed?

Side note: both googletest and google/benchmark use `cmath/math.h` but don't specify the `-lm` flag for FreeBSD/OpenBSD. There are additional linker errors that I'm also [addressing](https://github.com/google/benchmark/pull/1032#pullrequestreview-477425608) in those libraries that affects cctz on FreeBSD/OpenBSD.

devbww: > I do notice that building CCTZ using cmake on both FreeBSD and Linux produces output that shows a dependency on `libm`.

As does building using bazel on Linux.
```
$ cat bazel-bin/time_tool-2.params
-o
bazel-out/k8-fastbuild/bin/time_tool
-Wl,-S
-fuse-ld=gold
-Wl,-no-as-needed
-Wl,-z,relro,-z,now
-B/usr/bin
-pass-exit-codes
-lstdc++
-lm
bazel-out/k8-fastbuild/bin/_objs/time_tool/time_tool.pic.o
-Wl,--start-lib
bazel-out/k8-fastbuild/bin/_objs/time_zone/time_zone_fixed.pic.o
bazel-out/k8-fastbuild/bin/_objs/time_zone/time_zone_format.pic.o
bazel-out/k8-fastbuild/bin/_objs/time_zone/time_zone_if.pic.o
bazel-out/k8-fastbuild/bin/_objs/time_zone/time_zone_impl.pic.o
bazel-out/k8-fastbuild/bin/_objs/time_zone/time_zone_info.pic.o
bazel-out/k8-fastbuild/bin/_objs/time_zone/time_zone_libc.pic.o
bazel-out/k8-fastbuild/bin/_objs/time_zone/time_zone_lookup.pic.o
bazel-out/k8-fastbuild/bin/_objs/time_zone/time_zone_posix.pic.o
bazel-out/k8-fastbuild/bin/_objs/time_zone/zone_info_source.pic.o
-Wl,--end-lib
-Wl,--start-lib
bazel-out/k8-fastbuild/bin/_objs/civil_time/civil_time_detail.pic.o
-Wl,--end-lib
```
And that despite the CCTZ BUILD file not mentioning `-lm`, nor `-lstdc++`.

> And on FreeBSD/OpenBSD, for several projects, the `-lm` flag to the linker is usually explicit even if it is C++
source code. Here's an example from the [bazel source tree](https://github.com/bazelbuild/bazel/blob/ebf6a46e25d4be3bf0bbd77cb21dbff0276d5d04/src/tools/singlejar/BUILD#L76).

I can only ask the same question about that one ... why is the `-lm` linker option platform specific (unless the code has a platform-specific dependency on `math.h`)?

> Where do you reckon the flag would be more appropriately placed?

In this particular case, whoever generated the call to `ceilf()` should be the one to add the dependency.  In general, whoever creates the dependency should satisfy the dependency.

CCTZ has no idea that this dependency has been made, and having to be made aware of that indicates that something else is wrong.  `time_zone_impl.cc` has no `-lm` dependency, and in the Linux build neither does `time_zone_impl.pic.o`.  So, the BSD compiler must have introduced it, and therefore the BSD compiler should satisfy it.

> Side note: both googletest and google/benchmark use cmath/math.h but don't specify the -lm flag for FreeBSD/OpenBSD.

If a C++ program/library uses `<cmath>` and the cmath implementation uses libmath (which I assume it does), then the C++ compiler should add the `-lm` linker option (as is what I believed happened in the `time_tool` example above, despite nothing in `time_tool` using `<cmath>`).  [Can you build a simple `<cmath>`-using application on FreeBSD without specifying `-lm`?]

If a C++ program/library uses `<math.h>` then it should probably also arrange to link with `-lm`, but it probably gets by without doing so because the C++ compiler adds it anyway.

If a C program/library uses `<math.h>` then it should arrange to link with `-lm`.

yesudeep: You're right. Thank you for getting me to take a closer look at this. It turns out that Bazel on my system is using `/usr/bin/clang` instead of `/usr/bin/clang++` to compile C++ code. That's weird. I tried compiling the following with clang++ and g++, it worked fine. However, the C compiler frontend did complain.

It looks like this is the actual problem:

```shell
❯ /usr/bin/clang -o bazel-out/freebsd-fastbuild/bin/time_zone_lookup_test '-Wl,-rpath,$ORIGIN/_solib_freebsd/' -Lbazel-out/freebsd-fastbuild/bin/_solib_freebsd bazel-out/freebsd-fastbuild/bin/_objs/time_zone_lookup_test/time_zone_lookup_test.pic.o -llibtime_Uzone -llibcivil_Utime -lexternal_Scom_Ugoogle_Ugoogletest_Slibgtest_Umain -lexternal_Scom_Ugoogle_Ugoogletest_Slibgtest -pthread -Wl,-S -lstdc++ -Wl,-z,relro,-z,now -no-canonical-prefixes
ld: error: bazel-out/freebsd-fastbuild/bin/_solib_freebsd/liblibtime_Uzone.so: undefined reference to ceilf
clang: error: linker command failed with exit code 1 (use -v to see invocation)

❯ /usr/bin/clang++ -o bazel-out/freebsd-fastbuild/bin/time_zone_lookup_test '-Wl,-rpath,$ORIGIN/_solib_freebsd/' -Lbazel-out/freebsd-fastbuild/bin/_solib_freebsd bazel-out/freebsd-fastbuild/bin/_objs/time_zone_lookup_test/time_zone_lookup_test.pic.o -llibtime_Uzone -llibcivil_Utime -lexternal_Scom_Ugoogle_Ugoogletest_Slibgtest_Umain -lexternal_Scom_Ugoogle_Ugoogletest_Slibgtest -pthread -Wl,-S -lstdc++ -Wl,-z,relro,-z,now -no-canonical-prefixes
```

devbww: Yes, I believe that would explain it.  Thanks for digging.

[Note that you should also not have to specify `-lstdc++` explicitly on that `/usr/bin/clang++` command.]

I'll leave it to you to close the PR if you're happy that the problem lies elsewhere.

devbww:
For what it's worth, the CCTZ MacOS and Windows builds are currently failing with something like ...
```
external/com_github_google_benchmark/BUILD.bazel:13:1: no such target '@platforms//os:openbsd': target 'openbsd' not declared in package 'os' (did you mean 'freebsd'?) defined by external/platforms/os/BUILD and referenced by '@com_github_google_benchmark//:openbsd'
```
Perhaps it is due to https://github.com/google/benchmark/pull/1032/files.

yesudeep:
I'll take a look at it right away.

BUILD
yesudeep: Thank you! Sure, I'm learning quite a bit and have filed an issue here: https://github.com/bazelbuild/bazel/issues/12023

PR #169, opened 2020-09-10
Configures code to use clang-format's Google C++ style.

yesudeep:
Attempting to make it easier to submit consistently formatted code.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/169)


.clang-format
```diff
@@ -0,0 +1,168 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+
+# AlignConsecutiveAssignments: true
```
devbww, line 5: Let's remove all the comments, and just stick with `BasedOnStyle` and any exceptions.

.clang-format
```diff
@@ -0,0 +1,168 @@
+---
+Language: Cpp
```
devbww, line 2: Why do we need/want `Language`?

fmt
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
devbww, line 1: Let's remove this.  Even if we wanted to reformat en masse, it would only be useful once.

src/tzfile.h
```diff
@@ -22,36 +22,35 @@
 */
 
 #ifndef TZDIR
-#define TZDIR	"/usr/share/zoneinfo" /* Time zone object file directory */
-#endif /* !defined TZDIR */
+#define TZDIR "/usr/share/zoneinfo" /* Time zone object file directory */
```
devbww, line 25: This file is copied from another project (https://github.com/eggert/tz/blob/master/tzfile.h).  We shouldn't be reformatting it.

src/cctz_benchmark.cc
```diff
@@ -101,6 +101,7 @@ const char RFC3339_sec[] = "%Y-%m-%d%ET%H:%M:%S%Ez";
 const char RFC1123_full[] = "%a, %d %b %Y %H:%M:%S %z";
 const char RFC1123_no_wday[] = "%d %b %Y %H:%M:%S %z";
 
+// clang-format off
```
devbww, line 104: It seems preferable to put this inside the list, like the example in `src/civil_time_test.cc`.

src/time_tool.cc
```diff
@@ -34,6 +34,7 @@ template <typename D>
 using time_point = cctz::time_point<D>;
 using seconds = cctz::seconds;
 
+// clang-format off
```
devbww, line 37: Ditto.

src/time_zone_lookup_test.cc
```diff
@@ -32,6 +31,7 @@ namespace cctz {
 
 namespace {
 
+// clang-format off
```
devbww, line 34: Ditto.

devbww (commented):
Despite me saying we're OK with clang-format, I don't think we want to reformat everything en masse.  It just creates needless churn.

So, how about we limit things to the `.clang-format` file and the `clang-format on/off` directives for now.  Anything else can wait for if/when the actual offending block of code is modified.

.clang-format
yesudeep: This is the default configuration generated using the `clang-format` command. There are [other languages](https://clang.llvm.org/docs/ClangFormat.html
) clang-format formats too btw. I guess it keeps the configuration specific to C++.

.clang-format
yesudeep: Done.

.clang-format
yesudeep: Removed.

fmt
yesudeep: Done.

src/cctz_benchmark.cc
yesudeep: Done.

src/time_zone_lookup_test.cc
yesudeep: Done.

src/time_tool.cc
yesudeep: Done.

src/tzfile.h
yesudeep: Reverted.

yesudeep:
> Despite me saying we're OK with clang-format, I don't think we want to reformat everything en masse. It just creates needless churn.
> 
> So, how about we limit things to the `.clang-format` file and the `clang-format on/off` directives for now. Anything else can wait for if/when the actual offending block of code is modified.

Sure. I've limited the changes to only those files that contained clang-format guard directives. Any formatting around it is automatic as a result of the presence of the `.clang-format` configuration file.

src/cctz_benchmark.cc
```diff
@@ -104,6 +104,7 @@ const char RFC1123_no_wday[] = "%d %b %Y %H:%M:%S %z";
 // A list of known time-zone names.
 // TODO: Refactor with src/time_zone_lookup_test.cc.
 const char* const kTimeZoneNames[] = {
+    // clang-format off
```
devbww, line 107: Do you think we should indent the list another level to match the indentation of the comment/directive?

src/civil_time_test.cc
```diff
@@ -1022,6 +1022,7 @@ TEST(CivilTime, LeapYears) {
       int day;
     } leap_day;  // The date of the day after Feb 28.
   } kLeapYearTable[]{
+      // clang-format off
```
devbww, line 1025: For example, here they are aligned.

src/time_tool.cc
```diff
@@ -36,6 +36,7 @@ using seconds = cctz::seconds;
 
 // parse() specifiers for command-line time arguments.
 const char* const kFormats[] = {
+    // clang-format off
```
devbww, line 39: Ditto.

src/time_zone_lookup_test.cc
```diff
@@ -34,6 +33,7 @@ namespace {
 
 // A list of known time-zone names.
 const char* const kTimeZoneNames[] = {
+    // clang-format off
```
devbww, line 36: And again on the indentation.

src/time_tool.cc
```diff
@@ -85,13 +87,20 @@ bool ParseCivilSpec(const std::string& args, cctz::civil_second* when) {
...
 const char* WeekDayName(cctz::weekday wd) {
   switch (wd) {
-    case cctz::weekday::monday: return "Mon";
-    case cctz::weekday::tuesday: return "Tue";
-    case cctz::weekday::wednesday: return "Wed";
-    case cctz::weekday::thursday: return "Thu";
-    case cctz::weekday::friday: return "Fri";
-    case cctz::weekday::saturday: return "Sat";
-    case cctz::weekday::sunday: return "Sun";
+    case cctz::weekday::monday:
```
devbww, line 90: Let's drop all the other whitespace changes, unless you think the edit actually fixes something.

src/time_tool.cc
```diff
@@ -85,13 +87,15 @@ bool ParseCivilSpec(const std::string& args, cctz::civil_second* when) {
 
 const char* WeekDayName(cctz::weekday wd) {
   switch (wd) {
+    // clang-format off
```
devbww, line 90: I don't have a problem with what clang-format does here, so I don't want to disable it.  I'm just choosing not to make the edits until we have a need to otherwise modify these lines.  That is, the whitespace churn isn't worth it in isolation.  So, leave things be for now, unless there is some other good reason to make an edit (like marking those sections we know we don't want to be reflowed).

src/time_zone_lookup_test.cc
```diff
@@ -12,8 +12,6 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
-#include "cctz/time_zone.h"
```
devbww, line 15: This was here intentionally, so that we can test that `cctz/time_zone.h` includes what it needs.  That is, it is the "main include" (category 0).  I guess we'll have to do something to teach clang-format that, but it could wait.

src/time_zone_lookup_test.cc
```diff
@@ -813,6 +814,7 @@ TEST(BreakTime, TimePointResolution) {
   const time_zone utc = utc_time_zone();
   const auto t0 = chrono::system_clock::from_time_t(0);
 
+  // clang-format off
```
devbww, line 817: Ditto not wanting to defeat clang-format in cases like these (but not wanting to reformat them yet).

src/time_zone_lookup_test.cc
```diff
@@ -929,7 +931,7 @@ TEST(MakeTime, Normalization) {
 
 // NOTE: Run this with -ftrapv to detect overflow problems.
 TEST(MakeTime, SysSecondsLimits) {
-  const char RFC3339[] =  "%Y-%m-%d%ET%H:%M:%S%Ez";
+  const char RFC3339[] = "%Y-%m-%d%ET%H:%M:%S%Ez";
```
devbww, line 934: This looks like it may the only current example of unintended bad formatting.

devbww:
> I've limited the changes to only those files that contained clang-format guard directives. Any formatting around them is done automatically by the editor based on the .clang-format configuration file.

Just to clarify, I don't really want the whitespace churn in those files either.  Sorry.  I think you're going to have disable this automatic modification for this project.  I, for one, prefer to just clang-format individual statements as I write/modify them.

src/time_tool.cc
yesudeep: Okay. How do you propose to avoid formatting changes when editors/clang-format format the entire file as soon as one saves the file? The tools work at the file level. However, undoing these changes requires one to manually defeat clang-format, which appears to contradict the intention behind the guards and using `clang-format`? 

I'm not sure how defeating clang-format manually here helps maintain consistency. The idea was to format files so that any future feature-related changes would not have to include incidental formatting changes that were unrelated to the feature. Thoughts?

devbww: > How do you propose to avoid formatting changes when editors/clang-format format the entire file as soon as one saves the file?

I don't propose to do anything except reject changes that make whitespace-only edits to code they are otherwise not touching.

> The idea was to format files so that any future feature-related changes would not have to include/manually manage incidental formatting changes that were unrelated to the feature.

I agree that those future changes should not include incidental formatting changes.  They should just format the new/changed code as they see fit, and I'm happy for that to be as clang-format chooses (except in the few circumstances we've already discussed, hence the need for some extra controls).

> All the files were formatted to reduce churn, imo.

I guess I disagree.  Formatting for formatting's sake is the source of churn.

PR #174, opened 2020-10-19
Workaround gcc-5.5 ICE bug.

sbenzaquen:
Do not use mutating operations during constant evaluation of non-mutating operators to avoid triggering an ICE in gcc-5.5

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/174)

### DIFF
--- a/include/cctz/civil_time_detail.h
+++ b/include/cctz/civil_time_detail.h
@@ -382,16 +382,10 @@ class civil_time {
 
   // Assigning arithmetic.
   CONSTEXPR_M civil_time& operator+=(diff_t n) noexcept {
-    f_ = step(T{}, f_, n);
-    return *this;
+    return *this = *this + n;
   }
   CONSTEXPR_M civil_time& operator-=(diff_t n) noexcept {
-    if (n != (std::numeric_limits<diff_t>::min)()) {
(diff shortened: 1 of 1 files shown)


include/cctz/civil_time_detail.h
```diff
@@ -412,13 +406,17 @@ class civil_time {
...
   }
   friend CONSTEXPR_F civil_time operator+(diff_t n, civil_time a) noexcept {
-    return a += n;
+    return a + n;
   }
   friend CONSTEXPR_F civil_time operator-(civil_time a, diff_t n) noexcept {
-    return a -= n;
+    if (n != (std::numeric_limits<diff_t>::min)()) {
+      return civil_time(step(T{}, a.f_, -n));
+    } else {
```
devbww, line 417: We eschew `else` after `return`.  Care to reformulate?

devbww (commented):
> Do not use mutating operations during constant evaluation of non-mutating operators to avoid triggering an ICE

That's a shame ... I prefer the old incantation.

Thanks.

include/cctz/civil_time_detail.h
sbenzaquen: Done

devbww (approved):

PR #177, opened 2020-11-11
Fix tm_gmtoff build failure on AIX

### LINKED ISSUE #176: cctz fails to build on AIX
```
cctz/src/time_zone_libc.cc: In function 'bool cctz::{anonymous}::make_time(const civil_second&, int, time_t*, int*)':
cctz/src/time_zone_libc.cc:144:39: error: no matching function for call to 'tm_gmtoff(tm&)'
   *off = static_cast<int>(tm_gmtoff(tm));
                                       ^
cctz/src/time_zone_libc.cc:76:6: note: candidate: template<class T> decltype (tm.tm_gmtoff) cctz::{anonymous}::tm_gmtoff(const T&)
 auto tm_gmtoff(const T& tm) -> decltype(tm.tm_gmtoff) {
      ^~~~~~~~~
cctz/src/time_zone_libc.cc:76:6: note:   template argument deduction/substitution failed:
cctz/src/time_zone_libc.cc: In substitution of 'template<class T> decltype (tm.tm_gmtoff) cctz::{anonymous}::tm_gmtoff(const T&) [with T = tm]':
cctz/src/time_zone_libc.cc:144:39:   required from here
cctz/src/time_zone_libc.cc:76:44: error: 'const struct tm' has no member named 'tm_gmtoff'; did you mean 'tm_mon'?
 auto tm_gmtoff(const T& tm) -> decltype(tm.tm_gmtoff) {
                                         ~~~^~~~~~~~~
```

Full failure log: [R-lubridate-aix.txt](https://github.com/google/cctz/files/5521096/R-lubridate-aix.txt)

This is occurring because like Solaris, AIX does not provide a timezone offset field from `struct tm`. The Solaris code [here](https://github.com/tidyverse/lubridate/blob/master/src/cctz/src/time_zone_libc.cc#L92-L100) does work on AIX, it just needs a check for AIX as well. The only wrinkle is that AIX libc does export altzone, but does not define it in any header file, so an extern variable definition is needed.

kadler:
AIX does not provide a `struct tm` field for timezone offset, just like
Solaris so the existing code for that platform can be reused on AIX.
Unfortunately, AIX does not define the `altzone` variable in any header
file so a definition must be defined in this file.

Fixes #176

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/177)

### DIFF
--- a/src/time_zone_libc.cc
+++ b/src/time_zone_libc.cc
@@ -26,6 +26,12 @@
 #include "cctz/civil_time.h"
 #include "cctz/time_zone.h"
 
+#if defined(_AIX)
+extern "C" {
+  extern long altzone;
+}
+#endif
+
 namespace cctz {
 
 namespace {
@@ -40,7 +46,7 @@ auto tm_zone(const std::tm& tm) -> decltype(_tzname[0]) {
   const bool is_dst = tm.tm_isdst > 0;
   return _tzname[is_dst];
 }
-#elif defined(__sun)
+#elif defined(__sun) || defined(_AIX)
 // Uses the globals: 'timezone', 'altzone' and 'tzname'.
 auto tm_gmtoff(const std::tm& tm) -> decltype(timezone) {
   const bool is_dst = tm.tm_isdst > 0;


src/time_zone_libc.cc
```diff
@@ -26,6 +26,12 @@
 #include "cctz/civil_time.h"
 #include "cctz/time_zone.h"
 
+#ifdef _AIX
```
devbww, line 29: Nit: `#if defined(_AIX)` would be more consistent.

devbww (changes requested):
Thanks.  Just a couple of nits.

kadler:
I've adjusted the nits.

devbww (approved):
Thanks.

PR #178, opened 2020-12-08
add initial fuzzer for integration with OSS-Fuzz.

DavidKorczynski:
Hi maintainers,

It would be nice to have this project fuzzed by way of OSS-Fuzz, which is a service run by Google that performs continuous fuzzing of open source projects. 

Over here: https://github.com/google/oss-fuzz/pull/4806 I have set up the necessary artifacts to get the project integrated. In that PR you also see the fuzzer attached, but it would be nicer to have the fuzzer integrated into the upstream project itself rather than keeping it in OSS-Fuzz - so if this PR gets merged I will remove it from OSS-Fuzz. For integration, the only thing we need email of the maintainers that will be receiving access to bug reports, coverage reports and similar.

If you have any questions you can ask me, but as this is a Google open source project I assume you can reach out internally to @inferno-chromium

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/178)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,16 @@ if (BUILD_TESTING)
   )
 endif()
 
+if (BUILD_FUZZ)
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    add_compile_options(-fsanitize=fuzzer-no-link)
+    add_link_options(-fsanitize=fuzzer-no-link)
+  else()
+    message( SEND_ERROR "You must use clang++ for compiling fuzzers" )
+  endif()
+endif()
+
+
 # Starting from CMake >= 3.1, if a specific standard is required,
 # it can be set from the command line with:
 #     cmake -DCMAKE_CXX_STANDARD=[11|14|17]
@@ -102,6 +112,13 @@ set_target_properties(cctz PROPERTIES
   )
 target_link_libraries(cctz PUBLIC $<$<PLATFORM_ID:Darwin>:${CoreFoundation}>)  
 add_library(cctz::cctz ALIAS cctz)
+
+if (BUILD_FUZZ)
+  add_executable(cctz_fuzz src/cctz_fuzz.cc)
+  cctz_target_set_cxx_standard(cctz_fuzz)
+  target_link_libraries(cctz_fuzz cctz::cctz)
+  target_link_libraries(cctz_fuzz -fsanitize=fuzzer)
+endif()
 
 if (BUILD_TOOLS)
   add_executable(time_tool src/time_tool.cc)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ if (BUILD_FUZZ)
     add_compile_options(-fsanitize=fuzzer-no-link)
     add_link_options(-fsanitize=fuzzer-no-link)
   else()
-    message( SEND_ERROR "You must use clang++ for compiling fuzzers" )
+    message(SEND_ERROR "You must use clang++ for compiling fuzzers")
   endif()
 endif()
 
--- a/src/cctz_fuzz.cc
+++ b/src/cctz_fuzz.cc
@@ -1,0 +1,47 @@
+/* Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <fuzzer/FuzzedDataProvider.h>
+
+#include <iostream>
+#include <string>
+
+#include "cctz/civil_time.h"
+#include "cctz/time_zone.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	FuzzedDataProvider fuzzed_data(data, size);
+
+	cctz::time_zone lax;
+	std::string tz = fuzzed_data.ConsumeRandomLengthString();
+	if (load_time_zone(tz, &lax)) {
+		std::chrono::system_clock::time_point tp;
+		std::string date_format = fuzzed_data.ConsumeRandomLengthString();
+		std::string parse_format = fuzzed_data.ConsumeRandomLengthString();
+		cctz::parse(parse_format, date_format, lax, &tp);
+
+		const auto t1 = cctz::convert(cctz::civil_second(
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>()), lax);
+		std::string format = fuzzed_data.ConsumeRandomLengthString();
+		cctz::format(format, t1, lax);
+	}
+
+	return 0;
+}
--- a/src/cctz_fuzz.cc
+++ b/src/cctz_fuzz.cc
@@ -14,9 +14,6 @@
 
 #include <fuzzer/FuzzedDataProvider.h>
 
-#include <iostream>
-#include <string>
-
 #include <string>
 
 #include "cctz/civil_time.h"
--- a/src/cctz_fuzz.cc
+++ b/src/cctz_fuzz.cc
@@ -41,7 +41,8 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
                 fuzzed_data.ConsumeIntegral<std::uint64_t>(),
                 fuzzed_data.ConsumeIntegral<std::uint64_t>()), tz);
         std::string format = fuzzed_data.ConsumeRandomLengthString();
-        cctz::format(format, when, tz);
+        std::string formatted = cctz::format(format, when, tz);
+        cctz::parse(format, formatted, tz, &tp);
     }
 
     return 0;
--- a/src/cctz_fuzz.cc
+++ b/src/cctz_fuzz.cc
@@ -1,47 +1,48 @@
-/* Copyright 2020 Google LLC
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-      http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


src/cctz_fuzz.cc
```diff
@@ -0,0 +1,47 @@
...
+
+#include <fuzzer/FuzzedDataProvider.h>
+
+#include <iostream>
+#include <string>
+
+#include "cctz/civil_time.h"
+#include "cctz/time_zone.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
```
devbww, line 24: It would be nice if this used the same declaration style as the rest of the package, where we bind the `*` to the type rather than the name.  So, `const uint8_t* data`.

We also choose to `std` qualify the standard types, `std::uint8_t` and `std::size_t`.  But given that this is `extern "C"`, I guess there is some leeway there.

src/cctz_fuzz.cc
```diff
@@ -0,0 +1,47 @@
...
+	FuzzedDataProvider fuzzed_data(data, size);
+
+	cctz::time_zone lax;
+	std::string tz = fuzzed_data.ConsumeRandomLengthString();
+	if (load_time_zone(tz, &lax)) {
```
devbww, line 29: We choose not to rely on ADL, so `cctz::load_time_zone()`.

src/cctz_fuzz.cc
```diff
@@ -0,0 +1,47 @@
...
+#include <iostream>
+#include <string>
+
+#include "cctz/civil_time.h"
+#include "cctz/time_zone.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	FuzzedDataProvider fuzzed_data(data, size);
+
+	cctz::time_zone lax;
```
devbww, line 27: Why should this be called `lax` (a name commonly associated with America/Los_Angeles")?

src/cctz_fuzz.cc
```diff
@@ -0,0 +1,47 @@
...
+		std::string parse_format = fuzzed_data.ConsumeRandomLengthString();
+		cctz::parse(parse_format, date_format, lax, &tp);
+
+		const auto t1 = cctz::convert(cctz::civil_second(
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>()), lax);
```
devbww, line 41: s/uint32_t/std::uint32_t/

But why 32 bits?  In the internal version we use 64 bits, which is the type of arguments to `cctz::civil_second()` (with an allowance for year overflow)?

src/cctz_fuzz.cc
```diff
@@ -0,0 +1,47 @@
...
+
+	cctz::time_zone lax;
+	std::string tz = fuzzed_data.ConsumeRandomLengthString();
+	if (load_time_zone(tz, &lax)) {
+		std::chrono::system_clock::time_point tp;
+		std::string date_format = fuzzed_data.ConsumeRandomLengthString();
+		std::string parse_format = fuzzed_data.ConsumeRandomLengthString();
+		cctz::parse(parse_format, date_format, lax, &tp);
+
+		const auto t1 = cctz::convert(cctz::civil_second(
```
devbww, line 35: Why `t1` for a name?  It implies that there will probably be a `t2`.

src/cctz_fuzz.cc
```diff
@@ -0,0 +1,47 @@
...
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>(),
+				fuzzed_data.ConsumeIntegral<uint32_t>()), lax);
+		std::string format = fuzzed_data.ConsumeRandomLengthString();
+		cctz::format(format, t1, lax);
```
devbww, line 43: In the internal version we go on to exercise `cctz::parse()`.  Is there any reason not to do that here too?

src/cctz_fuzz.cc
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	return 0;
+}
```
devbww, line 47: How do we know this file even compiles?  To add it to this project we would at least have to check that locally somehow.

devbww (changes requested):

derekmauro:
We actually have a nearly identical program that we use for fuzzing internally. It may be better to just publish that program and add the necessary copybara changes internally.

DavidKorczynski:
@devbww and @derekmauro thanks for the comments. I am happy to apply the changes @devbww but if you have a fuzzer internally we may as well start with that one. I can then set up the necessary artifacts in the pull request on the OSS-Fuzz repository. Let me know if you prefer me to fix up the fuzzer according to the review or if you prefer to upload your own.

src/cctz_fuzz.cc
DavidKorczynski: I changed this to `when`, does that make more sense? I got the keyword from looking at other usages of `format` where the (old) `t1` is used.

src/cctz_fuzz.cc
DavidKorczynski: Added, is the current version what you had in mind?

src/cctz_fuzz.cc
DavidKorczynski: I set it up so you can build via cmake, e.g. `cmake -DBUILD_FUZZ=ON  -DBUILD_TESTING=OFF ../` - does this help with local testing?

DavidKorczynski:
@derekmauro @devbww do you have any updates on this one?

CMakeLists.txt
```diff
@@ -46,6 +46,16 @@ if (BUILD_TESTING)
   )
 endif()
 
+if (BUILD_FUZZ)
```
devbww, line 49: I'm no cmake expert, but should we `option(BUILD_FUZZ ...)` as well?

CMakeLists.txt
```diff
@@ -46,6 +46,16 @@ if (BUILD_TESTING)
...
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    add_compile_options(-fsanitize=fuzzer-no-link)
+    add_link_options(-fsanitize=fuzzer-no-link)
+  else()
+    message( SEND_ERROR "You must use clang++ for compiling fuzzers" )
```
devbww, line 54: Are the extra spaces next to the parentheses meaningful?

src/cctz_fuzz.cc
```diff
@@ -0,0 +1,49 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <fuzzer/FuzzedDataProvider.h>
+
+#include <iostream>
+#include <string>
+
+#include <string>
```
devbww, line 20: Duplicate `#include`?

src/cctz_fuzz.cc
```diff
@@ -0,0 +1,49 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <fuzzer/FuzzedDataProvider.h>
+
+#include <iostream>
```
devbww, line 17: Unused?

src/cctz_fuzz.cc
devbww: @derekmauro knows more about this, but as long as we can get that into the CI builds, it sounds like we'll at least be checking that it compiles.

src/cctz_fuzz.cc
devbww: That looks good, although I would call the date string `date_format` ... the latter implies it should contain format specifiers.

The internal version also simply uses a single format string for both `cctz::format()` and `cctz::parse()`.  I'm not sure if that helps, hurts, or doesn't matter.

devbww (commented):
> @derekmauro @devbww do you have any updates on this one?

Sorry for the delay.  I lost track of this over the holidays.

>> We actually have a nearly identical program that we use for fuzzing internally. It may be better to just publish that program and add the necessary copybara changes internally.
>
> Let me know if you prefer me to fix up the fuzzer according to the review or if you prefer to upload your own.

Sorry ... given the similarity between the two, I assumed you had the original version.

I'm all for merging the best parts of each, but I'm unclear on the logistics on copying internal code into GitHub.  @derekmauro may know more about what we can do there.

DavidKorczynski:
> > Sorry ... given the similarity between the two, I assumed you had the original version.
> I'm all for merging the best parts of each, but I'm unclear on the logistics on copying internal code into GitHub. @derekmauro may know more about what we can do there.

No problem, I don't have access to the internal version unfortunately. However, if you need assistance with the logistics of bringing internal fuzzers to Github then @inferno-chromium should be able to help you here. You should be able to reach him internally in Google.

derekmauro:
> > > Sorry ... given the similarity between the two, I assumed you had the original version.
> > > I'm all for merging the best parts of each, but I'm unclear on the logistics on copying internal code into GitHub. @derekmauro may know more about what we can do there.
> 
> No problem, I don't have access to the internal version unfortunately. However, if you need assistance with the logistics of bringing internal fuzzers to Github then @inferno-chromium should be able to help you here. You should be able to reach him internally in Google.

The hard part about this change is dealing with importing it, which is something I have to do. As mentioned, we already have this, we run it continuously, and the results are sent directly to me and a dedicated security team, so it isn't really a priority for me.  It may be best if I just opensource the existing thing, because the process for doing so is basically the same as importing this change. I just don't know when I'll get to it.

DavidKorczynski:
Ah fair enough - thanks @derekmauro and I understand that this is low priority then.

devbww:
OK, just to summarize:

- Maybe @DavidKorczynski can merge [google/oss-fuzz#4806](https://github.com/google/oss-fuzz/pull/4806), perhaps including the suggestions made here, in order to have something submitted to build from.
- In the meantime Google can continue to run their internal fuzzer.
- At some point, and at his convenience, @derekmauro will look into open-sourcing the best bits of those two versions into [google/cctz](https://github.com/google/cctz) (here), at which point the other two versions can probably disappear.

Thanks for everyone's help/input.

DavidKorczynski:
> OK, just to summarize:
> 
> * Maybe @DavidKorczynski can merge [google/oss-fuzz#4806](https://github.com/google/oss-fuzz/pull/4806), perhaps including the suggestions made here, in order to have something submitted to build from.
> * In the meantime Google can continue to run their internal fuzzer.
> * At some point, and at his convenience, @derekmauro will look into open-sourcing the best bits of those two versions into [google/cctz](https://github.com/google/cctz) (here), at which point the other two versions can probably disappear.
> 
> Thanks for everyone's help/input.

This works for me. If it works for @derekmauro then I can apply the final suggestions from this thread so this PR can get merged and the PR on oss-fuzz will get merged.

derekmauro:
@DavidKorczynski I don't think there is anything for you to do here with regards to this PR. We already have a nearly identical program (this program almost looks like it was written by the same person). When I get around to it, I'm likely just going to open source that.

DavidKorczynski:
> @DavidKorczynski I don't think there is anything for you to do here with regards to this PR. We already have a nearly identical program (this program almost looks like it was written by the same person). When I get around to it, I'm likely just going to open source that.

Sounds good. I will await signal for merging into OSS-Fuzz then (this PR https://github.com/google/oss-fuzz/pull/4806)

devbww:
There doesn't seem to be anything left to do within this PR.  Closing.

PR #185, opened 2021-03-29
Add support for cctz::load_time_zone("UTC0", &tz)

devbww:
`cctz::load_time_zone("UTC0", &tz)` now behaves exactly the same as
`cctz::load_time_zone("UTC", &tz)`, allowing `cctz::local_time_zone()`
to "succeed" (i.e., without fallback to "UTC") under `TZ=UTC0`.

This allows for setting `TZ=UTC0` to specify the local timezone as
UTC under both CCTZ and libraries that use POSIX-style `${TZ}` parsing.

Update `time_tool` to use an empty string to request `local_time_zone()`
instead of depending on `load_time_zone("localtime", &tz)` doing the
same thing, which may not work with the configuration of some systems.
This means that `time_tool` now respects `${TZ}`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/185)

### DIFF
--- a/src/time_zone_fixed.cc
+++ b/src/time_zone_fixed.cc
@@ -48,11 +48,9 @@ int Parse02d(const char* p) {
 }  // namespace
 
 bool FixedOffsetFromName(const std::string& name, seconds* offset) {
-  if (name.rfind("UTC", 0, 3) == 0) {
-    if (name.size() == 3 || (name.size() == 4 && name[3] == '0')) {
-      *offset = seconds::zero();  // "UTC" or "UTC0"
-      return true;
-    }
+  if (name == "UTC" || name == "UTC0") {
+    *offset = seconds::zero();
+    return true;
   }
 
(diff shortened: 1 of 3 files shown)


src/time_zone_fixed.cc
```diff
@@ -48,9 +48,11 @@ int Parse02d(const char* p) {
 }  // namespace
 
 bool FixedOffsetFromName(const std::string& name, seconds* offset) {
-  if (name.compare(0, std::string::npos, "UTC", 3) == 0) {
-    *offset = seconds::zero();
-    return true;
+  if (name.rfind("UTC", 0, 3) == 0) {
```
devjgm, line 51: What do you think about simply: `if (name == "UTC" || name == "UTC0")`?

devjgm (approved):

src/time_zone_fixed.cc
devbww: Sure.

Issue #187, opened 2021-04-06
Help with underflows on Fuchsia


kpozin:
I'm trying to get cctz's tests passing on Fuchsia, currently targeting x64 (arm64 is next).

The last part of `MakeTime.SysSecondsLimits` in `time_zone_lookup_test.cc` is failing:
```text
Expected equality of these values:
  "-2147481748-01-01T00:00:00+00:00"
  format(RFC3339, tp, utc)
    Which is: "-9223372036854775808-01-01T00:00:00+00:00"
```

Setting a year of `-2147481748` apparently causes an underflow. Where in the cctz library (or somewhere else?) should I be looking to fix this behavior?

Thanks!

devbww:
It looks like the expectation you are referring to is ...
```
EXPECT_EQ("-2147481748-01-01T00:00:00+00:00", format(RFC3339, tp, cut));
```
around line 1025.  If so, my first concern is that the test has been changed because that line uses the timezone `cut`, whereas your output says `utc`.

That said, that part of the test is about conversions in the **libc:UTC** timezone, which means it mostly concerns the behavior of `gmtime_r(3)` from your C library.

It is difficult to say more without being about to run the code, but my first question would be _"how does `gmtime_t()` behave on a `time_t` value of -67768040609740800?"_  My guess is that it fails.  It should return `{-2147483648, 0, 1, 0, 0, 0}`.

devbww:
Did you discover anything about the behavior of your `gmtime_t()`, or is this no longer an issue?

kpozin:
I apologize for not getting back to this earlier.

Fuchsia uses musl for Posix-related libraries:
* [gmtime_r.c](https://cs.opensource.google/fuchsia/fuchsia/+/master:zircon/third_party/ulib/musl/src/time/gmtime_r.c)
* [__secs_to_tm.c](https://cs.opensource.google/fuchsia/fuchsia/+/master:zircon/third_party/ulib/musl/src/time/__secs_to_tm.c)

`gmtime_r()` for `-67768040609740800` does indeed fail; it yields a `tm` with all fields set to 0.

These are the two underflow conditions in `__secs_to_tm`:
```cpp
  if (t < INT_MIN * 31622400LL || t > INT_MAX * 31622400LL)
    return -1;
```
```
  if (years + 100 > INT_MAX || years + 100 < INT_MIN)
    return -1;
```
The second is the one that's triggered, with an off-by-one error. (The year ends up being  `-2147483649`.)

[...]

PR #189, opened 2021-06-08
Add basic support for Fuchsia

kpozin:
- Implement `FuchsiaZoneInfoSource` to read tzif2 files from their designated locations on Fuchsia systems.
- Implement calls to Fuchsia APIs in `local_time_zone()`.
- Disable a time conversion edge case that is unsupported on Fuchsia.

Bazel does not yet support Fuchsia configs, so these changes must be built and tested internally. See cl/374587662.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/189)

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -39,15 +39,11 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <fstream>
 #include <functional>
 #include <memory>
 #include <sstream>
 #include <string>
-#include <utility>
-
-#if defined(__Fuchsia__)
-#include <fstream>
-#endif
 
 #include "cctz/civil_time.h"
 #include "time_zone_fixed.h"
@@ -56,6 +52,12 @@
 namespace cctz {
 
 namespace {
+
+#if defined(__Fuchsia__)
+const bool kIsFuchsia = true;
+#else
+const bool kIsFuchsia = false;
+#endif
 
 inline bool IsLeap(year_t year) {
   return (year % 4) == 0 && ((year % 100) != 0 || (year % 400) == 0);
@@ -606,6 +608,22 @@ inline FILE* FOpen(const char* path, const char* mode) {
 #endif
 }
 
+// Returns the size (in bytes) of the given file.
+std::size_t FSize(FILE* fp) {
+  if (fp == nullptr) {
+    return 0;
+  }
+  std::size_t length = 0;
+  if (fseek(fp, 0, SEEK_END) == 0) {
+    long offset = ftell(fp);
+    if (offset >= 0) {
+      length = static_cast<std::size_t>(offset);
+    }
+    rewind(fp);
+  }
+  return length;
+}
+
 // A stdio(3)-backed implementation of ZoneInfoSource.
 class FileZoneInfoSource : public ZoneInfoSource {
  public:
@@ -632,11 +650,6 @@ class FileZoneInfoSource : public ZoneInfoSource {
   explicit FileZoneInfoSource(
       FILE* fp, std::size_t len = std::numeric_limits<std::size_t>::max())
       : fp_(fp, fclose), len_(len) {}
-
-  // Returns a file pointer and the size of the file in bytes. If the file
-  // cannot be opened, returns a null file pointer and a size of 0.
-  static std::pair<FILE*, std::size_t> OpenZoneInfoFile(
-      const std::string& path);
 
  private:
   std::unique_ptr<FILE, int(*)(FILE*)> fp_;
@@ -668,27 +681,10 @@ std::unique_ptr<ZoneInfoSource> FileZoneInfoSource::Open(
   path.append(name, pos, std::string::npos);
 
   // Open the zoneinfo file.
-  auto fp_and_len = FileZoneInfoSource::OpenZoneInfoFile(path);
-  if (fp_and_len.first == nullptr) {
-    return nullptr;
-  }
-  return std::unique_ptr<ZoneInfoSource>(
-      new FileZoneInfoSource(fp_and_len.first, fp_and_len.second));
-}
-
-std::pair<FILE*, std::size_t> FileZoneInfoSource::OpenZoneInfoFile(
-    const std::string& path) {
   FILE* fp = FOpen(path.c_str(), "rb");
-  if (fp == nullptr) return std::make_pair(nullptr, 0);
-  std::size_t length = 0;
-  if (fseek(fp, 0, SEEK_END) == 0) {
-    long offset = ftell(fp);
-    if (offset >= 0) {
-      length = static_cast<std::size_t>(offset);
-    }
-    rewind(fp);
-  }
-  return std::make_pair(fp, length);
+  if (fp == nullptr) return nullptr;
+  std::size_t length = FSize(fp);
+  return std::unique_ptr<ZoneInfoSource>(new FileZoneInfoSource(fp, length));
 }
 
 class AndroidZoneInfoSource : public FileZoneInfoSource {
@@ -745,8 +741,7 @@ std::unique_ptr<ZoneInfoSource> AndroidZoneInfoSource::Open(
   return nullptr;
 }
 
-#if defined(__Fuchsia__)
-// An info source for using cctz inside Fuchsia components. This attempts to
+// A zoneinfo source for use inside Fuchsia components. This attempts to
 // read zoneinfo files from one of several known paths in a component's
 // incoming namespace. [Config data][1] is preferred, but package-specific
 // resources are also supported.
@@ -766,13 +761,15 @@ class FuchsiaZoneInfoSource : public FileZoneInfoSource {
   std::string version_;
 };
 
-// Reads the contents of the file as ASCII. Returns an empty string on error.
-std::string ReadFileAsString(const std::string& path) {
+// Reads the first line of the file into a string. Returns an empty string on
+// error.
+std::string ReadFirstLine(const std::string& path) {
+  std::string line;
   std::ifstream input_stream(path);
-  if (!input_stream.good()) return "";
-  std::string contents((std::istreambuf_iterator<char>(input_stream)),
-                       std::istreambuf_iterator<char>());
-  return contents;
+  if (input_stream.good()) {
+    std::getline(input_stream, line);
+  }
+  return line;
 }
 
 std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
@@ -793,34 +790,36 @@ std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
 
   // Use of the "file:" prefix is intended for testing purposes only.
   const std::size_t pos = (name.compare(0, 5, "file:") == 0) ? 5 : 0;
-  const bool is_test_path = (pos != name.size() && name[pos] == '/');
+  const bool is_absolute_path = (pos != name.size() && name[pos] == '/');
   const auto path_prefixes =
-      is_test_path ? kEmptyPathPrefixes : kTzdataPathPrefixes;
+      is_absolute_path ? kEmptyPathPrefixes : kTzdataPathPrefixes;
 
   for (const char* tzdata_dir : path_prefixes) {
     // Map the time-zone name to a path name.
     // Fuchsia builds place time zone files at <tzdata>/<format>/...
     std::string path = tzdata_dir;
-    if (!is_test_path) {
+    if (!is_absolute_path) {
       path += kZoneinfoFormatPrefix;
     }
     path.append(name, pos, std::string::npos);
 
-    const auto fp_and_len = FileZoneInfoSource::OpenZoneInfoFile(path);
-    if (fp_and_len.first == nullptr) {
+    FILE* fp = FOpen(path.c_str(), "rb");
+    if (fp == nullptr) {
       continue;
     }
 
+    std::size_t len = FSize(fp);
     std::string version_path = tzdata_dir;
     version_path += kVersionFileName;
-    std::string version_str = ReadFileAsString(version_path);
+    // revision.txt should contain no newlines, but to be defensive we read just
+    // the first line.
+    std::string version_str = ReadFirstLine(version_path);
 
     return std::unique_ptr<ZoneInfoSource>(new FuchsiaZoneInfoSource(
-        fp_and_len.first, fp_and_len.second, version_str.c_str()));
+        fp, len, version_str.c_str()));
   }
   return nullptr;
 }
-#endif
 }  // namespace
 
 bool TimeZoneInfo::Load(const std::string& name) {
@@ -836,12 +835,11 @@ bool TimeZoneInfo::Load(const std::string& name) {
   // Find and use a ZoneInfoSource to load the named zone.
   auto zip = cctz_extension::zone_info_source_factory(
       name, [](const std::string& n) -> std::unique_ptr<ZoneInfoSource> {
-#if defined(__Fuchsia__)
-        if (auto z = FuchsiaZoneInfoSource::Open(n)) return z;
-#else
+        if (kIsFuchsia) {
+          if (auto z = FuchsiaZoneInfoSource::Open(n)) return z;
+        }
         if (auto z = FileZoneInfoSource::Open(n)) return z;
         if (auto z = AndroidZoneInfoSource::Open(n)) return z;
-#endif
         return nullptr;
       });
   return zip != nullptr && Load(zip.get());
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -53,6 +53,12 @@ namespace cctz {
 
 namespace {
 
+#if defined(__ANDROID__) || defined(ANDROID)
+const bool kIsAndroid = true;
+#else
+const bool kIsAndroid = false;
+#endif
+
 #if defined(__Fuchsia__)
 const bool kIsFuchsia = true;
 #else
@@ -608,22 +614,6 @@ inline FILE* FOpen(const char* path, const char* mode) {
 #endif
 }
 
-// Returns the size (in bytes) of the given file.
-std::size_t FSize(FILE* fp) {
-  if (fp == nullptr) {
-    return 0;
-  }
-  std::size_t length = 0;
-  if (fseek(fp, 0, SEEK_END) == 0) {
-    long offset = ftell(fp);
-    if (offset >= 0) {
-      length = static_cast<std::size_t>(offset);
-    }
-    rewind(fp);
-  }
-  return length;
-}
-
 // A stdio(3)-backed implementation of ZoneInfoSource.
 class FileZoneInfoSource : public ZoneInfoSource {
  public:
@@ -683,7 +673,14 @@ std::unique_ptr<ZoneInfoSource> FileZoneInfoSource::Open(
   // Open the zoneinfo file.
   FILE* fp = FOpen(path.c_str(), "rb");
   if (fp == nullptr) return nullptr;
-  std::size_t length = FSize(fp);
+  std::size_t length = 0;
+  if (fseek(fp, 0, SEEK_END) == 0) {
+    long offset = ftell(fp);
+    if (offset >= 0) {
+      length = static_cast<std::size_t>(offset);
+    }
+    rewind(fp);
+  }
   return std::unique_ptr<ZoneInfoSource>(new FileZoneInfoSource(fp, length));
 }
 
@@ -755,8 +752,8 @@ class FuchsiaZoneInfoSource : public FileZoneInfoSource {
   std::string Version() const override { return version_; }
 
  private:
-  explicit FuchsiaZoneInfoSource(FILE* fp, std::size_t len, const char* vers)
-      : FileZoneInfoSource(fp, len), version_(vers) {}
+  explicit FuchsiaZoneInfoSource(FILE* fp, const char* vers)
+      : FileZoneInfoSource(fp), version_(vers) {}
 
   std::string version_;
 };
@@ -808,15 +805,14 @@ std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
       continue;
     }
 
-    std::size_t len = FSize(fp);
     std::string version_path = tzdata_dir;
     version_path += kVersionFileName;
     // revision.txt should contain no newlines, but to be defensive we read just
     // the first line.
     std::string version_str = ReadFirstLine(version_path);
 
-    return std::unique_ptr<ZoneInfoSource>(new FuchsiaZoneInfoSource(
-        fp, len, version_str.c_str()));
+    return std::unique_ptr<ZoneInfoSource>(
+        new FuchsiaZoneInfoSource(fp, version_str.c_str()));
   }
   return nullptr;
 }
@@ -835,11 +831,13 @@ bool TimeZoneInfo::Load(const std::string& name) {
   // Find and use a ZoneInfoSource to load the named zone.
   auto zip = cctz_extension::zone_info_source_factory(
       name, [](const std::string& n) -> std::unique_ptr<ZoneInfoSource> {
+        if (auto z = FileZoneInfoSource::Open(n)) return z;
+        if (kIsAndroid) {
+          if (auto z = AndroidZoneInfoSource::Open(n)) return z;
+        }
         if (kIsFuchsia) {
           if (auto z = FuchsiaZoneInfoSource::Open(n)) return z;
         }
-        if (auto z = FileZoneInfoSource::Open(n)) return z;
-        if (auto z = AndroidZoneInfoSource::Open(n)) return z;
         return nullptr;
       });
   return zip != nullptr && Load(zip.get());
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -43,6 +43,11 @@
 #include <memory>
 #include <sstream>
 #include <string>
+#include <utility>
+
+#if defined(__Fuchsia__)
+#include <fstream>
+#endif
 
 #include "cctz/civil_time.h"
 #include "time_zone_fixed.h"
@@ -628,6 +633,11 @@ class FileZoneInfoSource : public ZoneInfoSource {
       FILE* fp, std::size_t len = std::numeric_limits<std::size_t>::max())
       : fp_(fp, fclose), len_(len) {}
 
+  // Returns a file pointer and the size of the file in bytes. If the file
+  // cannot be opened, returns a null file pointer and a size of 0.
+  static std::pair<FILE*, std::size_t> OpenZoneInfoFile(
+      const std::string& path);
+
  private:
   std::unique_ptr<FILE, int(*)(FILE*)> fp_;
   std::size_t len_;
@@ -658,8 +668,18 @@ std::unique_ptr<ZoneInfoSource> FileZoneInfoSource::Open(
   path.append(name, pos, std::string::npos);
 
   // Open the zoneinfo file.
+  auto fp_and_len = FileZoneInfoSource::OpenZoneInfoFile(path);
+  if (fp_and_len.first == nullptr) {
+    return nullptr;
+  }
+  return std::unique_ptr<ZoneInfoSource>(
+      new FileZoneInfoSource(fp_and_len.first, fp_and_len.second));
+}
+
+std::pair<FILE*, std::size_t> FileZoneInfoSource::OpenZoneInfoFile(
+    const std::string& path) {
   FILE* fp = FOpen(path.c_str(), "rb");
-  if (fp == nullptr) return nullptr;
+  if (fp == nullptr) return std::make_pair(nullptr, 0);
   std::size_t length = 0;
   if (fseek(fp, 0, SEEK_END) == 0) {
     long offset = ftell(fp);
@@ -668,7 +688,7 @@ std::unique_ptr<ZoneInfoSource> FileZoneInfoSource::Open(
     }
     rewind(fp);
   }
-  return std::unique_ptr<ZoneInfoSource>(new FileZoneInfoSource(fp, length));
+  return std::make_pair(fp, length);
 }
 
 class AndroidZoneInfoSource : public FileZoneInfoSource {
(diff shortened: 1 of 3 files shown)


kpozin:
CC @filmil

filmil (commented):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/cctz/189)* status: all files reviewed, 7 unresolved discussions (waiting on @kpozin)

___
*[src/time_zone_info.cc, line 638 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnJJMa87SRckKPdvOV:-MbnJJMb4J1d2SkOEf3X:b-mpmjlu) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L638)):*
> ```cpp
>   // Returns a file pointer and the size of the file in bytes. If the file
>   // cannot be opened, returns a null file pointer and a size of 0.
>   static std::pair<FILE*, std::size_t> OpenZoneInfoFile(
> ```

Who closes the FILE*?
___
*[src/time_zone_info.cc, line 671 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnJeDcEtcxEiNXG338:-MbnJeDcEtcxEiNXG339:b-dntol5) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L671)):*
> ```cpp
> 
>   // Open the zoneinfo file.
>   auto fp_and_len = FileZoneInfoSource::OpenZoneInfoFile(path);
> ```

May wanna unpack with `std::tie` to avoid needing to use `.first` and `.second` which aren't too informative.  C++ version permitting.
___
*[src/time_zone_info.cc, line 679 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKD672iqSsmCE4OFQ:-MbnKD672iqSsmCE4OFR:bpoqqe0) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L679)):*
> ```cpp
> }
> 
> std::pair<FILE*, std::size_t> FileZoneInfoSource::OpenZoneInfoFile(
> ```

Is there any way to avoid returning a naked `FILE*` here?   It is all too easy to forget to close, or worse, leak the file if say an exception happens.
___
*[src/time_zone_info.cc, line 759 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKmjnBftXIW5I7NId:-MbnKmjnBftXIW5I7NIe:b79d43g) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L759)):*
> ```cpp
> class FuchsiaZoneInfoSource : public FileZoneInfoSource {
>  public:
>   static std::unique_ptr<ZoneInfoSource> Open(const std::string& name);
> ```

Document public API params?
___
*[src/time_zone_info.cc, line 763 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKbdE-rFNmQLi71rQ:-MbnKbdE-rFNmQLi71rR:b-ocxqx5) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L763)):*
> ```cpp
> 
>  private:
>   explicit FuchsiaZoneInfoSource(FILE* fp, std::size_t len, const char* vers)
> ```

I wonder, is there any way to use the C++ I/O here - `FILE*` gives me pause?
___
*[src/time_zone_lookup.cc, line 159 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnLJEq37JPpPzryijb:-MbnLJEq37JPpPzryijc:b2ws88d) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L159)):*
> ```cpp
>     std::unique_ptr<sys::ComponentContext> context =
>         sys::ComponentContext::Create();
>     fuchsia::intl::PropertyProviderPtr intl_provider =
> ```

Would a `SyncPtr` be simpler here, by any chance? You are effectively synchronizing on the call returning.
___
*[src/time_zone_lookup_test.cc, line 1026 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnLYhD2GL118qapFPP:-MbnLYhD2GL118qapFPQ:bb69o2h) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup_test.cc#L1026)):*
> ```cpp
>     tp = convert(civil_second(min_tm_year, 1, 1, 0, 0, 0), cut);
> #if defined(__Fuchsia__)
>     // Fuchsia's gmtime_r() fails on extreme negative tm_year values.
> ```

#til

devbww (changes requested):
> Bazel does not yet support Fuchsia configs, so these changes must be
> tested internally. See cl/374587662.

We probably shouldn't be mentioning "internally" here, or giving links to inaccessible information.  This is an open-source project.

Not being able to build/test this code in CCTZ comes close to a fatal flaw too.  There is no way I can guarantee that future changes won't break Fuchsia if we can't support something like a Kokoro build here.  The same thing is currently true for Android, but see the discussion about conditional compilation.

> *[Reviewable](https://reviewable.io/reviews/google/cctz/189#-:-Mbnn46sAjaq_6tFGe2A:b5knka2)* status: all files reviewed, 15 unresolved discussions (waiting on @filmil and @kpozin)

___
*[src/time_zone_info.cc, line 638 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnJJMa87SRckKPdvOV:-Mbnoa-v5tpc5yR0o5k6:b-sijcdm) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L638)):*
<details><summary><i>Previously, filmil (Filip Filmar) wrote…</i></summary><blockquote>

Who closes the FILE*?
</blockquote></details>

The `FileZoneInfoSource` dtor, via the `FileZoneInfoSource::fp_` dtor.  But, see the discussion below.
___
*[src/time_zone_info.cc, line 679 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKD672iqSsmCE4OFQ:-Mbnq5aLAL2TWlNFqXd_:b-6af4pk) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L679)):*
<details><summary><i>Previously, filmil (Filip Filmar) wrote…</i></summary><blockquote>

Is there any way to avoid returning a naked `FILE*` here?   It is all too easy to forget to close, or worse, leak the file if say an exception happens.
</blockquote></details>

Previously the disposition of the `FILE*` was much clearer ... straight from the result of `FOpen()` to the `FileZoneInfoSource` ctor.  The imposition of the `pair` adds an extra layer.

And the `pair` is pretty much what a `FileZoneInfoSource` is, together with the closing property you ask about.  So, I would think that there is someway to avoid the `pair` altogether.

Perhaps one simple approach is to factor out a `std::size_t FSize(FILE*)` helper.  Then `FileZoneInfoSource::Open()` and `FuchsiaZoneInfoSource::Open()` can both just call `FOpen()` and `FSize()` independently, and there is no need for a `pair` at all.

That would also get rid of the `OpenZoneInfoFile()` function, which actually had nothing to do with "zoneinfo".
___
*[src/time_zone_info.cc, line 749 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo2EHP8ICaS6ci0WBd:-Mbo2EHP8ICaS6ci0WBe:b608l7z) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L749)):*
> ```cpp
> 
> #if defined(__Fuchsia__)
> // An info source for using cctz inside Fuchsia components. This attempts to
> ```

I would suggest `s/info/zoneinfo/`.  We don't use "info" anywhere else.

I'd also suggest rewording to avoid "using cctz".  This is all about using cctz.  (If we have to refer to the package in other documentation we tend to say "CCTZ".)
___
*[src/time_zone_info.cc, line 759 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKmjnBftXIW5I7NId:-Mbnw37d8rkViKaiqbXo:b4hpnex) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L759)):*
<details><summary><i>Previously, filmil (Filip Filmar) wrote…</i></summary><blockquote>

Document public API params?
</blockquote></details>

This isn't public.  The inter-module documentation is on `TimeZoneIf::Load()`, and the actual public documentation is on `cctz::load_time_zone()`.
___
*[src/time_zone_info.cc, line 769 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbny_AD4E4LBpjw8pbd:-Mbny_AD4E4LBpjw8pbe:bm729fr) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L769)):*
> ```cpp
> };
> 
> // Reads the contents of the file as ASCII. Returns an empty string on error.
> ```

This doesn't seem to have anything to do with ASCII.
___
*[src/time_zone_info.cc, line 796 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnzTWUBXZ4qlQ4yV29:-MbnzTWUBXZ4qlQ4yV2A:b-ep93wc) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L796)):*
> ```cpp
>   // Use of the "file:" prefix is intended for testing purposes only.
>   const std::size_t pos = (name.compare(0, 5, "file:") == 0) ? 5 : 0;
>   const bool is_test_path = (pos != name.size() && name[pos] == '/');
> ```

I'm not sure what "test" means, but `is_test_path` seems better as `is_absolute_path`.  Then not adding a prefix nor a format prefix to an absolute path makes more sense.
___
*[src/time_zone_info.cc, line 816 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnypCPE25Fbhc9TDkF:-MbnypCPE25Fbhc9TDkG:bqd9m99) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L816)):*
> ```cpp
>     std::string version_path = tzdata_dir;
>     version_path += kVersionFileName;
>     std::string version_str = ReadFileAsString(version_path);
> ```

Does the "revision.txt" file have a terminating newline?  If so, you'll want to omit it from `version_str`.  (You may also want to be defensive about data beyond one line.)
___
*[src/time_zone_info.cc, line 839 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo0U1y7vfJJuuQl4jz:-Mbo0SKz0MEktovJyWH7:bxzdaqz) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L839)):*
> ```cpp
>   auto zip = cctz_extension::zone_info_source_factory(
>       name, [](const std::string& n) -> std::unique_ptr<ZoneInfoSource> {
> #if defined(__Fuchsia__)
> ```

I would suggest that every platform should support finding zoneinfo files in `${TZDIR}`.  That is, that `FileZoneInfoSource::Open() is unconditionally first.

We can discuss whether anything after that needs to be conditional (at the cost of some failed opens).  Notice that this module supports Android, but doesn't conditionally compile anything for Android.  This helps with avoiding Android breakages when we don't build/test for it.
___
*[src/time_zone_lookup.cc, line 151 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo7Oxs465hUJOoTEhK:-Mbo7Oxs465hUJOoTEhL:bajmn10) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L151)):*
> ```cpp
> #endif
> #if defined(__Fuchsia__)
>   // Note: Declared outside of unnamed scope to avoid early destruction.
> ```

I'd drop this note.  Everything is declared where it needs to be.
___
*[src/time_zone_lookup.cc, line 154 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo9K5JBFxjKPGG94Pm:-Mbo9K5JBFxjKPGG94Pn:bopvji0) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L154)):*
> ```cpp
>   std::string primary_tz;
>   {
>     const zx::duration kTimeout = zx::msec(500);
> ```

I don't understand your architecture of course, but is timing out a real possibility?  Perhaps this is related to Filip's `SyncPtr` question.
___
*[src/time_zone_lookup.cc, line 159 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo7nvDCDgii1w2OJnm:-Mbo7nvDCDgii1w2OJnn:b-7t4fp1) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L159)):*
> ```cpp
>     std::unique_ptr<sys::ComponentContext> context =
>         sys::ComponentContext::Create();
>     fuchsia::intl::PropertyProviderPtr intl_provider =
> ```

auto?
___
*[src/time_zone_lookup_test.cc, line 1026 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnLYhD2GL118qapFPP:-Mbo8Ico7lDbrhYEWZOM:b-8qqn5q) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup_test.cc#L1026)):*
<details><summary><i>Previously, filmil (Filip Filmar) wrote…</i></summary><blockquote>

#til
</blockquote></details>

See my responses on https://github.com/google/cctz/issues/187, where I noted:

It is difficult to say more without being able to run the code, but my first question would be "how does gmtime_r() behave on a `time_t` value of -67768040609740800?" My guess is that it fails. It should return `{0, 0, 0, 1, 0, -2147483648}`.

Which C library do you use?

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/google/cctz/189)* status: all files reviewed, 15 unresolved discussions (waiting on @filmil and @kpozin)

___
*[src/time_zone_info.cc, line 679 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKD672iqSsmCE4OFQ:-MbogoGZ5eia57vCv5M2:bdj907u) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L679)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

Previously the disposition of the `FILE*` was much clearer ... straight from the result of `FOpen()` to the `FileZoneInfoSource` ctor.  The imposition of the `pair` adds an extra layer.

And the `pair` is pretty much what a `FileZoneInfoSource` is, together with the closing property you ask about.  So, I would think that there is someway to avoid the `pair` altogether.

Perhaps one simple approach is to factor out a `std::size_t FSize(FILE*)` helper.  Then `FileZoneInfoSource::Open()` and `FuchsiaZoneInfoSource::Open()` can both just call `FOpen()` and `FSize()` independently, and there is no need for a `pair` at all.

That would also get rid of the `OpenZoneInfoFile()` function, which actually had nothing to do with "zoneinfo".
</blockquote></details>

That said, there is actually no need to predetermine the length of the file.  The `len` argument to the `FileZoneInfoSource` ctor can be omitted when the file is simply TZif.  The `len` argument is only needed in cases like `AndroidZoneInfoSource`, where the TZif data has been encapsulated inside another file.

[That begs the question of why `FileZoneInfoSource::Open()` bothers to specify `len`, but that is either an historical artifact or a refactoring error.  I'll remove it.]

The bottom line is that `FuchsiaZoneInfoSource::Open()` can just use `FOpen()` alone (if I understand your file layout).

kpozin (commented):
We're willing to live with the possibility of being broken upstream and submitting retroactive fixes for now; Fuchsia is already in this position with a few other open-source libraries.

I removed the conditional compilation where possible, but we're stuck with it in `time_zone_lookup.cc`, which requires Fuchsia-only libraries.

> *[Reviewable](https://reviewable.io/reviews/google/cctz/189#-:-MbsI2t88SKULiIyeZzX:b-5x7058)* status: 0 of 3 files reviewed, 15 unresolved discussions (waiting on @devbww and @filmil)

___
*[src/time_zone_info.cc, line 638 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnJJMa87SRckKPdvOV:-MbsIYHl-vq_o7GIbb65:bb6sh2i) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L638)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

The `FileZoneInfoSource` dtor, via the `FileZoneInfoSource::fp_` dtor.  But, see the discussion below.
</blockquote></details>

Ack.
___
*[src/time_zone_info.cc, line 671 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnJeDcEtcxEiNXG338:-MbsI_Jt98DCYzgl3ONM:bhufgfa) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L671)):*
<details><summary><i>Previously, filmil (Filip Filmar) wrote…</i></summary><blockquote>

May wanna unpack with `std::tie` to avoid needing to use `.first` and `.second` which aren't too informative.  C++ version permitting.
</blockquote></details>

I switched to @devbww's suggested refactoring.
___
*[src/time_zone_info.cc, line 679 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKD672iqSsmCE4OFQ:-MbrHmut31UUfSzcgYS5:b7ucoll) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L679)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

That said, there is actually no need to predetermine the length of the file.  The `len` argument to the `FileZoneInfoSource` ctor can be omitted when the file is simply TZif.  The `len` argument is only needed in cases like `AndroidZoneInfoSource`, where the TZif data has been encapsulated inside another file.

[That begs the question of why `FileZoneInfoSource::Open()` bothers to specify `len`, but that is either an historical artifact or a refactoring error.  I'll remove it.]

The bottom line is that `FuchsiaZoneInfoSource::Open()` can just use `FOpen()` alone (if I understand your file layout).
</blockquote></details>

`std::size_t FSize(FILE*)` sounds like the best option. I'll keep `FileZoneInfoSource` as the base class of `FuchsiaZoneInfoSource` for now, so `len` is still needed for the constructor.
___
*[src/time_zone_info.cc, line 749 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo2EHP8ICaS6ci0WBd:-MbrJFAeDNQXDHYg4Y6e:b-896fix) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L749)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

I would suggest `s/info/zoneinfo/`.  We don't use "info" anywhere else.

I'd also suggest rewording to avoid "using cctz".  This is all about using cctz.  (If we have to refer to the package in other documentation we tend to say "CCTZ".)
</blockquote></details>

Done.
___
*[src/time_zone_info.cc, line 759 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKmjnBftXIW5I7NId:-MbrJG-G0xb2sZyoxDNU:b2nhb2c) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L759)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

This isn't public.  The inter-module documentation is on `TimeZoneIf::Load()`, and the actual public documentation is on `cctz::load_time_zone()`.
</blockquote></details>

Ack
___
*[src/time_zone_info.cc, line 763 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKbdE-rFNmQLi71rQ:-MbrJLzU3MgzUNPonp6C:b2nhb2c) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L763)):*
<details><summary><i>Previously, filmil (Filip Filmar) wrote…</i></summary><blockquote>

I wonder, is there any way to use the C++ I/O here - `FILE*` gives me pause?
</blockquote></details>

Ack
___
*[src/time_zone_info.cc, line 769 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbny_AD4E4LBpjw8pbd:-MbrEdNJ4cUt2YPHpiT-:bjfzbd7) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L769)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

This doesn't seem to have anything to do with ASCII.
</blockquote></details>

Corrected comment.
___
*[src/time_zone_info.cc, line 796 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnzTWUBXZ4qlQ4yV29:-MbrFu_-5s8mqX7VZJiS:b-896fix) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L796)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

I'm not sure what "test" means, but `is_test_path` seems better as `is_absolute_path`.  Then not adding a prefix nor a format prefix to an absolute path makes more sense.
</blockquote></details>

Done.
___
*[src/time_zone_info.cc, line 816 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnypCPE25Fbhc9TDkF:-MbrJkCAC_HMGmSSwvAj:b-3k5z4o) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L816)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

Does the "revision.txt" file have a terminating newline?  If so, you'll want to omit it from `version_str`.  (You may also want to be defensive about data beyond one line.)
</blockquote></details>

Done. No newlines in the format, but I'll use the defensive approach.
___
*[src/time_zone_info.cc, line 839 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo0U1y7vfJJuuQl4jz:-MbrMpWKAtN-llqbvA_E:b-hasdlw) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L839)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

I would suggest that every platform should support finding zoneinfo files in `${TZDIR}`.  That is, that `FileZoneInfoSource::Open() is unconditionally first.

We can discuss whether anything after that needs to be conditional (at the cost of some failed opens).  Notice that this module supports Android, but doesn't conditionally compile anything for Android.  This helps with avoiding Android breakages when we don't build/test for it.
</blockquote></details>

What if we added a `const bool kIsFuchsia` based on `defined(__Fuchsia__`? Then the compilation will be unconditional but we won't make extraneous calls at runtime.
___
*[src/time_zone_lookup.cc, line 151 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo7Oxs465hUJOoTEhK:-MbrNOW6CHdH-nFnS8ZV:b-896fix) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L151)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

I'd drop this note.  Everything is declared where it needs to be.
</blockquote></details>

Done.
___
*[src/time_zone_lookup.cc, line 154 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo9K5JBFxjKPGG94Pm:-MbrNPNz1XcKeV4lDD3X:bfjl9ka) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L154)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

I don't understand your architecture of course, but is timing out a real possibility?  Perhaps this is related to Filip's `SyncPtr` question.
</blockquote></details>

Added an explanatory comment to the code.
Yes, there are some failure states where the method would never return unless there was a timeout. The synchronous API doesn't currently allow timeouts.
___
*[src/time_zone_lookup.cc, line 159 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnLJEq37JPpPzryijb:-MbrNte_0fd0h-PbvDuU:b-h44w9t) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L159)):*
<details><summary><i>Previously, filmil (Filip Filmar) wrote…</i></summary><blockquote>

Would a `SyncPtr` be simpler here, by any chance? You are effectively synchronizing on the call returning.
</blockquote></details>

(See above. The sync API doesn't allow timeouts.)
___
*[src/time_zone_lookup.cc, line 159 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo7nvDCDgii1w2OJnm:-MbrO5wi0uoJgSV1yZmj:b-896fix) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L159)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

auto?
</blockquote></details>

Done.
___
*[src/time_zone_lookup_test.cc, line 1026 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnLYhD2GL118qapFPP:-MbrO9MT5FUkOHvF9YSi:bmgewna) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup_test.cc#L1026)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

See my responses on https://github.com/google/cctz/issues/187, where I noted:

It is difficult to say more without being able to run the code, but my first question would be "how does gmtime_r() behave on a `time_t` value of -67768040609740800?" My guess is that it fails. It should return `{0, 0, 0, 1, 0, -2147483648}`.

Which C library do you use?
</blockquote></details>

We're using musl. I just responded on #187 with a rough analysis of the bug in their `gmtime_r` implementation, but fixing musl's date calculations is a bit out of my scope for now :)

devbww (changes requested):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/cctz/189)* status: all files reviewed, 9 unresolved discussions (waiting on @filmil and @kpozin)

___
*[src/time_zone_info.cc, line 679 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKD672iqSsmCE4OFQ:-Mbt--KO21tIvTmjNIjL:b-q64pkt) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L679)):*
<details><summary><i>Previously, kpozin (Konstantin Pozin) wrote…</i></summary><blockquote>

`std::size_t FSize(FILE*)` sounds like the best option. I'll keep `FileZoneInfoSource` as the base class of `FuchsiaZoneInfoSource` for now, so `len` is still needed for the constructor.
</blockquote></details>

See above.
___
*[src/time_zone_info.cc, line 839 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo0U1y7vfJJuuQl4jz:-Mbt-WGj6yFoZo0hbffM:b-cfpwt3) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L839)):*
<details><summary><i>Previously, kpozin (Konstantin Pozin) wrote…</i></summary><blockquote>

What if we added a `const bool kIsFuchsia` based on `defined(__Fuchsia__`? Then the compilation will be unconditional but we won't make extraneous calls at runtime.
</blockquote></details>

That's fine I guess, although I'm not sure I would worry about the failing opens in the non-Fuchsia cases.

However, I still think the `FileZoneInfoSource::Open()` should always come first.
___
*[src/time_zone_info.cc, line 625 at r2](https://reviewable.io/reviews/google/cctz/189#-Mbsyzu-DUF0P04AgxSK:-Mbsyzu-DUF0P04AgxSL:b-r0zi1y) ([raw file](https://github.com/google/cctz/blob/448c5b1f940ef98411df6135197c1c02397a51d4/src/time_zone_info.cc#L625)):*
> ```cpp
>   }
>   return length;
> }
> ```

To clarify what I was suggesting, `FSize()` is not actually needed.  In the two cases it is currently used, the second argument to the `FileZoneInfoSource` ctor can instead be omitted (giving the parameter its default value, which will allow reading to EOF).

That is, delete this function, remove its use from `FileZoneInfoSource::Open()`, and remove `len` from your new `FuchsiaZoneInfoSource` ctor.

devbww (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/cctz/189)* status: all files reviewed, 9 unresolved discussions (waiting on @filmil and @kpozin)

___
*[src/time_zone_info.cc, line 625 at r2](https://reviewable.io/reviews/google/cctz/189#-Mbsyzu-DUF0P04AgxSK:-MbwNGPb3npUAIb1gWXP:b-5gns5v) ([raw file](https://github.com/google/cctz/blob/448c5b1f940ef98411df6135197c1c02397a51d4/src/time_zone_info.cc#L625)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

To clarify what I was suggesting, `FSize()` is not actually needed.  In the two cases it is currently used, the second argument to the `FileZoneInfoSource` ctor can instead be omitted (giving the parameter its default value, which will allow reading to EOF).

That is, delete this function, remove its use from `FileZoneInfoSource::Open()`, and remove `len` from your new `FuchsiaZoneInfoSource` ctor.
</blockquote></details>

I do have a pending edit to clean the file-size stuff up independently, if you would like me to get that in first.

kpozin (commented):
> *[Reviewable](https://reviewable.io/reviews/google/cctz/189)* status: 2 of 3 files reviewed, 9 unresolved discussions (waiting on @devbww and @filmil)

___
*[src/time_zone_info.cc, line 679 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnKD672iqSsmCE4OFQ:-MbwDJPt1Wf48o0GJY5t:b-896fix) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L679)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

See above.
</blockquote></details>

Done.
___
*[src/time_zone_info.cc, line 839 at r1](https://reviewable.io/reviews/google/cctz/189#-Mbo0U1y7vfJJuuQl4jz:-MbwPfxzEFJ0G_6kTyNP:bjwgxfk) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_info.cc#L839)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

That's fine I guess, although I'm not sure I would worry about the failing opens in the non-Fuchsia cases.

However, I still think the `FileZoneInfoSource::Open()` should always come first.
</blockquote></details>

Done. Also added a `const bool kIsAndroid`.
___
*[src/time_zone_info.cc, line 625 at r2](https://reviewable.io/reviews/google/cctz/189#-Mbsyzu-DUF0P04AgxSK:-MbwDDMYBEdrYneAZiLI:bn203y8) ([raw file](https://github.com/google/cctz/blob/448c5b1f940ef98411df6135197c1c02397a51d4/src/time_zone_info.cc#L625)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

I do have a pending edit to clean the file-size stuff up independently, if you would like me to get that in first.
</blockquote></details>

Done. Sorry for my misunderstanding.

devbww (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/cctz/189)* status: all files reviewed, 7 unresolved discussions (waiting on @filmil)

filmil (commented):
> *[Reviewable](https://reviewable.io/reviews/google/cctz/189)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kpozin)

___
*[src/time_zone_lookup.cc, line 159 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnLJEq37JPpPzryijb:-Mbwjwa954mzdFQffZRE:bxumkec) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup.cc#L159)):*
<details><summary><i>Previously, kpozin (Konstantin Pozin) wrote…</i></summary><blockquote>

(See above. The sync API doesn't allow timeouts.)
</blockquote></details>

OK

Separately, I wonder if we can do something about the need to watch for a timeout.
___
*[src/time_zone_lookup_test.cc, line 1026 at r1](https://reviewable.io/reviews/google/cctz/189#-MbnLYhD2GL118qapFPP:-MbwkBqy5az4hQZN5cTV:bechevw) ([raw file](https://github.com/google/cctz/blob/e8e74cdbf8c0d108f390cf9a55e55a230f601b35/src/time_zone_lookup_test.cc#L1026)):*
<details><summary><i>Previously, kpozin (Konstantin Pozin) wrote…</i></summary><blockquote>

We're using musl. I just responded on #187 with a rough analysis of the bug in their `gmtime_r` implementation, but fixing musl's date calculations is a bit out of my scope for now :)
</blockquote></details>

OK

Closing this comment, I'm OK with what kpozin@ said.  Feel free to reopen if you think otherwise.

kpozin:
I think we're all set. Please merge whenever convenient.

PR #190, opened 2021-06-14
Move raw FILE* values into an auto-closing unique_ptr ASAP.

devbww:
All `FileZoneInfoSource`-derived classes now accept transfer of
of that `unique_ptr`.

Also remove the `FileZoneInfoSource::Open()` code to determine
the length of the zoneinfo file.  We do not need the length when
we read to EOF.

Finally, refactor the new `FuchsiaZoneInfoSource` code a little.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/190)

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -771,7 +771,7 @@ std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
   const auto prefixes = name_absolute ? kEmptyPrefix : kTzdataPrefixes;
 
   // Fuchsia builds place zoneinfo files at "<prefix><format><name>".
-  for (const std::string prefix : prefixes) {
+  for (const std::string& prefix : prefixes) {
     std::string path = prefix;
     if (!prefix.empty()) path += "zoneinfo/tzif2/";  // format
     path.append(name, pos, std::string::npos);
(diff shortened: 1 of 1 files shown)


src/time_zone_info.cc
```diff
@@ -752,70 +749,54 @@ class FuchsiaZoneInfoSource : public FileZoneInfoSource {
...
-    std::string path = tzdata_dir;
-    if (!is_absolute_path) {
-      path += kZoneinfoFormatPrefix;
-    }
+  const auto kEmptyPrefix = {""};
+  const bool name_absolute = (pos != name.size() && name[pos] == '/');
+  const auto prefixes = name_absolute ? kEmptyPrefix : kTzdataPrefixes;
+
+  // Fuchsia builds place zoneinfo files at "<prefix><format><name>".
+  for (const std::string prefix : prefixes) {
```
devjgm, line 774: `prefix` could be a const-ref here. I don't think it needs to be a value.

devjgm (approved):

src/time_zone_info.cc
devbww: I don't think it makes a difference either way, as in each case we're just constructing a `std::string` from a `const char*`.  I usually prefer to give that constructed value a name rather than binding a reference to a temporary, but OK.

PR #191, opened 2021-06-24
Fixes -Wrange-loop-construct warning in GCC 11

derekmauro:
This is similar to #188, but fixes the issue introduced in #190.

I'm sending a separate change to upgrade to GCC 11 to prevent these
from sneaking in.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/191)

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -771,7 +771,7 @@ std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
   const auto prefixes = name_absolute ? kEmptyPrefix : kTzdataPrefixes;
 
   // Fuchsia builds place zoneinfo files at "<prefix><format><name>".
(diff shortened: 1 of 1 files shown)


src/time_zone_info.cc
```diff
@@ -771,7 +771,7 @@ std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
   const auto prefixes = name_absolute ? kEmptyPrefix : kTzdataPrefixes;
 
   // Fuchsia builds place zoneinfo files at "<prefix><format><name>".
-  for (const std::string& prefix : prefixes) {
+  for (const std::string prefix : prefixes) {
```
devbww, line 774: Ergh.  See the unresolved conversation in #190.  That is, I blame Greg!  :-)

devbww (approved):
Thanks.

PR #193, opened 2021-07-11
CMake: fix cross-build to iOS when tools are enabled

SpaceIm:
1. When we cross-build to iOS, by default PLATFORM_ID is not set, while APPLE variable is. Without this modification, CoreFoundation is not linked under iOS/tvOS/watchOS, leading to undefined references (at link time of dependent tools).

2. Due to https://cmake.org/cmake/help/latest/policy/CMP0006.html, RUNTIME DESTINATION for executables is not sufficient for iOS/tvOS/watchOS, they need BUNDLE DESTINATION.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/193)


devbww (approved):
Thanks.

PR #194, opened 2021-07-11
CMake: fix installation of dll on windows

SpaceIm:
It fixes installation of dll on Windows.

For MinGW it's fine, all symbols are exported by default,.
But for msvc, symbols are not exported, so shared is broken (no import lib created). The quick & dirty solution is WINDOWS_EXPORT_ALL_SYMBOLS in CMakeLists.txt, but it requires CMake 3.4, and current min version of cctz is 2.8, therefore I've decided to not add it (unless you allow to bump min CMake version).

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/194)


devbww (commented):
Thanks.

> The quick & dirty solution is WINDOWS_EXPORT_ALL_SYMBOLS in CMakeLists.txt, but it requires CMake 3.4, and current min version of cctz is 2.8, therefore I've decided to not add it (unless you allow to bump min CMake version).

We're not opposed to bumping the minimum CMake version, but not being CMake experts means we don't understand how that is usually handled.

Is there some way to use `WINDOWS_EXPORT_ALL_SYMBOLS` if >= 3.4 is in use?

SpaceIm:
Sure, `WINDOWS_EXPORT_ALL_SYMBOLS` works for all CMake versions greater than 3.4:

```cmake
cmake_minimum_required(VERSION 3.4)

(...)

set_property(TARGET cctz PROPERTY WINDOWS_EXPORT_ALL_SYMBOLS ON)
```

devbww (commented):
Sorry for not being clearer, but I was wondering about something built upon `if (CMAKE_VERSION VERSION_LESS "3.4")`, where we could deal with the issue in different ways depending on the facilities available.

> *[Reviewable](https://reviewable.io/reviews/google/cctz/194#-:-MeRn1Dd8hPlz4tu3dWW:b8dm0v3)* status: 0 of 1 files reviewed, all discussions resolved

SpaceIm:
Well, you can add `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)`, even if CMake min version is less than 3.4, it won't "fail" if you use CMake < 3.4, but it won't create an import library if Visual Studio and shared.
I don't recommend branching based on CMake version, it makes CMakeLists cumbersome and hard to maintain.

devbww (approved):
> Well, you can add set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON), even if CMake min version is less than 3.4, it won't "fail" if you use CMake < 3.4, but it won't create an import library if Visual Studio and shared.

I was envisaging an additional, similar conditional on the `RUNTIME DESTINATION` change here, but ...

> I don't recommend branching based on CMake version, it makes CMakeLists cumbersome and hard to maintain.

OK (although note that we already have a few).

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/cctz/194#-:-MeW5ov55Mj35aX0hIeM:bgar61s)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SpaceIm)

PR #205, opened 2021-10-19
Remove a few redundant calls to `unique_ptr.get()`

dinord:
These trigger clang tidy checks when importing into google3: https://clang.llvm.org/extra/clang-tidy/checks/readability-redundant-smartptr-get.html

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/205)


devbww (approved):
Thanks.

PR #208, opened 2021-11-01
Fix 32-bit ARM builds

derekmauro:
This patch changes the linking of 32-bit ARM zone info source to use
the same code as the 32-bit X86 zone info source.

This change is requested in
https://github.com/abseil/abseil-cpp/issues/829 by members of the
Microsoft organization who say it fixes the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/208)


devbww (approved):
Thanks @derekmauro.

PR #213, opened 2022-01-19
[Fuchsia] Reduce number of threads for LoadZonesConcurrently.

Colemalban:
This reduces the thread count from 128 -> 64 when running on Fuchsia.
With the higher thread count, the test was flaky and would often time
out.

To test, I ran this 100 times to verify that it passes each time.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/213)

### DIFF
--- a/src/time_zone_lookup_test.cc
+++ b/src/time_zone_lookup_test.cc
@@ -685,7 +685,14 @@ TEST(TimeZones, LoadZonesConcurrently) {
     }
   };
 
+#if defined(__Fuchsia__)
+  // On Fuchsia, the test will often timeout with the thread count set to 128.
+  // This lower thread count ensures the test passes consistently.
+  const std::size_t n_threads = 64;
+#else
   const std::size_t n_threads = 128;
+#endif
+
   std::vector<std::thread> threads;
   std::vector<std::set<std::string>> thread_failures(n_threads);
   for (std::size_t i = 0; i != n_threads; ++i) {
--- a/src/time_zone_lookup_test.cc
+++ b/src/time_zone_lookup_test.cc
@@ -685,14 +685,7 @@ TEST(TimeZones, LoadZonesConcurrently) {
     }
   };
 
-#if defined(__Fuchsia__)
-  // On Fuchsia, the test will often timeout with the thread count set to 128.
-  // This lower thread count ensures the test passes consistently.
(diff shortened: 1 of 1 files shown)


src/time_zone_lookup_test.cc
```diff
@@ -685,7 +685,14 @@ TEST(TimeZones, LoadZonesConcurrently) {
     }
   };
 
+#if defined(__Fuchsia__)
+  // On Fuchsia, the test will often timeout with the thread count set to 128.
+  // This lower thread count ensures the test passes consistently.
+  const std::size_t n_threads = 64;
```
devbww, line 691: > With the higher thread count, the test was flaky and would often time out.

Is this implying that the test would succeed if the test timeout was increased?  What is the current timeout?

What value does `std::thread::hardware_concurrency()` return on your platform/test-environment?  Might we be better off using that unconditionally?

Colemalban: I believe the current timeout is 300 seconds. Increasing the timeout could cause the test to not be flaky, but there is no guarantee.  I'm also not sure what value would achieve that.

Using std::thread::hardware_concurrency does seem to be another good option. On Fuchsia emulators it returns 4 and on my workstation (non-fuchsia config) it returns 112. Both work fine with the test. 

My most recent commit uses std::thread::hardware_concurrency.

devbww: I'm still concerned that there is some underlying issue here that should not be ignored.  Even with only 4 cores, do we expect loading all the timezones (perhaps up to 64 times) to take over 300 seconds?  Perhaps so, and in which case I'm with you that "perhaps up to 64 times" is too many.

src/time_zone_lookup_test.cc
```diff
@@ -685,7 +685,7 @@ TEST(TimeZones, LoadZonesConcurrently) {
     }
   };
 
-  const std::size_t n_threads = 128;
+  const std::size_t n_threads = std::thread::hardware_concurrency();
```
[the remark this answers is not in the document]
devbww: We should probably also guard against _"If the value is not well defined or not computable, returns ​0​."_

devbww:
Is this still an issue?

devbww:
> Is this still an issue?

Apparently not.  Closing for now.

Issue #212, opened 2022-01-18
loadtimezone and n_sec impl is slow in starrocks


aaawuanjun:
other olap use cctz, it occure 2 function inside cctz
n_sec and loadTimeZone
it calls too much time, so kernel goes the spin_lock_slowpath

![image](https://user-images.githubusercontent.com/35732071/149874352-41fd6648-7c56-4185-b5e4-77f5d386b9f0.png)

devbww:
I'm afraid your report lacks enough context so as to render it unactionable.

If you believe any CCTZ functions are doing unnecessary work, or have ideas about optimizations, please submit your detailed thoughts.  Thanks.

aaawuanjun:
First, change normal lock to readwrite lock in LoadTimeZone method

aaawuanjun:
about read write lock, see pull #214 

before modify, lock has race condition
online information:
![image](https://user-images.githubusercontent.com/35732071/150304522-9e3b59ca-59b1-43d0-9e2e-0cb3486bf8b4.png)

test env information:
![image](https://user-images.githubusercontent.com/35732071/150305176-3fdf2400-1740-4bb8-949c-9df13ecf8cda.png)

after this, lock's hot will decrease
test env information:
![image](https://user-images.githubusercontent.com/35732071/150306526-0ee1d679-46e0-49ac-b5ae-97262b097136.png)

 brpc's lock info inside LoadTimeZone  will down

aaawuanjun:
what's the period of new version comes out?

PR #214, opened 2022-01-20
change normal lock to readwrite lock in LoadTimeZone func

aaawuanjun:
issue #212

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/214)

### DIFF
--- a/src/time_zone_impl.cc
+++ b/src/time_zone_impl.cc
@@ -32,8 +32,8 @@ using TimeZoneImplByName =
     std::unordered_map<std::string, const time_zone::Impl*>;
 TimeZoneImplByName* time_zone_map = nullptr;
 
-typedef std::shared_lock<std::shared_mutex> ReadLock;
-typedef std::lock_guard<std::shared_mutex> WriteLock;
+typedef std::shared_lock<std::shared_mutex> read_lock;
+typedef std::lock_guard<std::shared_mutex> write_lock;
 
 // Mutual exclusion for time_zone_map.
 std::shared_mutex& TimeZoneMutex() {
@@ -61,7 +61,7 @@ bool time_zone::Impl::LoadTimeZone(const std::string& name, time_zone* tz) {
 
   // Check whether the time zone has already been loaded.
   {
-    ReadLock lock(TimeZoneMutex());
+    read_lock lock(TimeZoneMutex());
     if (time_zone_map != nullptr) {
       TimeZoneImplByName::const_iterator itr = time_zone_map->find(name);
       if (itr != time_zone_map->end()) {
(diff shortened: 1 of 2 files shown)


aaawuanjun:
compile env requires:
g++ version needs bigger than 7.x, I use gcc version: 10.x
image

src/time_zone_impl.cc
```diff
@@ -58,7 +61,7 @@ bool time_zone::Impl::LoadTimeZone(const std::string& name, time_zone* tz) {
 
   // Check whether the time zone has already been loaded.
   {
-    std::lock_guard<std::mutex> lock(TimeZoneMutex());
+    read_lock lock(TimeZoneMutex());
```
devbww, line 64: If you are seeing contention on this mutex, which is only held over an `unordered_map` lookup, then I can only suggest that you are calling `cctz::load_time_zone()` way too much.  Typically you would only load a time zone once, and then use the returned `cctz::time_zone` value after that.  But if you need to build your own `string`->`cctz::time_zone` cache, that you can tune to your own needs, so be it.

src/time_zone_impl.cc
```diff
@@ -32,11 +32,14 @@ using TimeZoneImplByName =
...
+typedef std::shared_lock<std::shared_mutex> read_lock;
+typedef std::lock_guard<std::shared_mutex> write_lock;
+
 // Mutual exclusion for time_zone_map.
-std::mutex& TimeZoneMutex() {
+std::shared_mutex& TimeZoneMutex() {
   // This mutex is intentionally "leaked" to avoid the static deinitialization
   // order fiasco (std::mutex's destructor is not trivial on many platforms).
-  static std::mutex* time_zone_mutex = new std::mutex;
+  static std::shared_mutex* time_zone_mutex = new std::shared_mutex;
```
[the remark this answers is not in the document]
devbww: > I will attempt to get our CI builds to enforce that restriction.

It looks like they already do ... at least on Linux and macOS.

src/time_zone_impl.cc
aaawuanjun: we made a cache in [OLAP]starrocks(https://github.com/StarRocks/starrocks/pull/2990), to avoid calling cctz inner functions too much times, at the same time, we also want to optimize lock in cctz.

aaawuanjun: about the lock of optimize, https://github.com/StarRocks/starrocks/pull/2966

devbww: I'm sure it is easier for you to build a high-performance cache given the platform and application knowledge of your app than it is for me in a general-purpose library using only C++11.  So, good luck.

[Aside:  I would wonder about the model behind anything that uses the term "timezone_offsets".]

PR #225, opened 2022-08-02
Make cctz civil(fields) constructor public

sbruceheart:
This allows use of a faster constructor when it is certain that provided year/month/day/hour/minute/second are within acceptable ranges.

---
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/225)


sbruceheart:
Incorrect push, apologies.

PR #237, opened 2022-11-02
build: fix cross-platform android builds on OS X

alyssawilk:
Fixing https://github.com/abseil/abseil-cpp/issues/326 by removing a non-working Foundation linkopt for OS X and iOS.
Foundation should now be implicitly added by modern versions of bazel so this will hopefully not be problematic for builds previously affected by https://github.com/abseil/abseil-cpp/issues/291

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/237)

co-authored-by: [devbww](https://github.com/devbww)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -14,13 +14,6 @@
 
 licenses(["notice"])
 
-config_setting(
-    name = "ios",
-    constraint_values = [
-        "@platforms//os:ios",
-    ],
-)
-
 ### libraries
 
 cc_library(
@@ -59,12 +52,9 @@ cc_library(
         "include/cctz/zone_info_source.h",
     ],
     includes = ["include"],
-    linkopts = select({
-        "//:ios": [
-            "-framework Foundation",
-        ],
-        "//conditions:default": [],
-    }),
+    # Note that OS X and iOS both have an implicit dependency on Foundation but
+    # it is no longer explicitly added as a linkopt as bazel adds it
+    # automatically. See https://github.com/abseil/abseil-cpp/issues/326 for details.
     visibility = ["//visibility:public"],
     deps = [":civil_time"],
 )
--- a/BUILD
+++ b/BUILD
@@ -52,9 +52,10 @@ cc_library(
         "include/cctz/zone_info_source.h",
     ],
     includes = ["include"],
-    # Note that OS X and iOS both have an implicit dependency on Foundation but
-    # it is no longer explicitly added as a linkopt as bazel adds it
-    # automatically. See https://github.com/abseil/abseil-cpp/issues/326 for details.
+    # Note that OS X and iOS both have a dependency on Foundation but it is no
+    # longer explicitly added as a linkopt as that causes problems cross-compiling android
+    # builds on OSX, and bazel will add it automatically.
+    # See https://github.com/abseil/abseil-cpp/issues/326 for details.
     visibility = ["//visibility:public"],
     deps = [":civil_time"],
 )
--- a/BUILD
+++ b/BUILD
@@ -67,9 +67,6 @@ cc_library(
     ],
     includes = ["include"],
     linkopts = select({
-        "//:osx": [
-            "-framework Foundation",
-        ],
         "//:ios": [
             "-framework Foundation",
         ],
--- a/BUILD
+++ b/BUILD
@@ -15,13 +15,6 @@
 licenses(["notice"])
 
 config_setting(
-    name = "osx",
-    constraint_values = [
-        "@platforms//os:osx",
-    ],
-)
-
-config_setting(
     name = "ios",
     constraint_values = [
         "@platforms//os:ios",
--- a/BUILD
+++ b/BUILD
@@ -52,10 +52,9 @@ cc_library(
         "include/cctz/zone_info_source.h",
     ],
     includes = ["include"],
-    # Note that OS X and iOS both have a dependency on CoreFoundation but it is
-    # no longer explicitly added as a linkopt as that causes problems
-    # cross-compiling android builds on OSX, and bazel will add it
-    # automatically.
+    # OS X and iOS no longer use `linkopts = ["-framework CoreFoundation"]`
+    # as (1) bazel adds it automatically, and (2) it caused problems when
+    # cross-compiling for Android.
     # See https://github.com/abseil/abseil-cpp/issues/326 for details.
     visibility = ["//visibility:public"],
     deps = [":civil_time"],
--- a/BUILD
+++ b/BUILD
@@ -52,9 +52,10 @@ cc_library(
         "include/cctz/zone_info_source.h",
     ],
     includes = ["include"],
-    # Note that OS X and iOS both have a dependency on Foundation but it is no
-    # longer explicitly added as a linkopt as that causes problems cross-compiling android
-    # builds on OSX, and bazel will add it automatically.
+    # Note that OS X and iOS both have a dependency on CoreFoundation but it is
+    # no longer explicitly added as a linkopt as that causes problems
+    # cross-compiling android builds on OSX, and bazel will add it
+    # automatically.
     # See https://github.com/abseil/abseil-cpp/issues/326 for details.
     visibility = ["//visibility:public"],
     deps = [":civil_time"],


BUILD
```diff
@@ -67,9 +67,6 @@ cc_library(
     ],
     includes = ["include"],
     linkopts = select({
-        "//:osx": [
-            "-framework Foundation",
-        ],
```
devbww, line 72: This was added in https://github.com/google/cctz/pull/102, which references https://github.com/abseil/abseil-cpp/issues/291.

Do we understand why that was wrong, or what has changed?

BUILD
```diff
@@ -67,9 +67,6 @@ cc_library(
...
         "//:ios": [
             "-framework Foundation",
         ],
```
devbww, line 72: Does any of the same logic apply to "ios"?

BUILD
```diff
@@ -67,9 +67,6 @@ cc_library(
     ],
     includes = ["include"],
     linkopts = select({
-        "//:osx": [
```
devbww, line 70: If this goes, the matching `config_setting()` can too.

devbww (commented):
Perhaps the PR title could be more descriptive of the actual change.

BUILD
alyssawilk: This hasn't caused a problem for Envoy Mobile, and we do iOS builds

devjgm (commented):
Thanks for the PR, Alyssa. Can you tell me how to repro the error you're seeing? I just checked out cctz on my Mac and built it w/ `bazel test ...` and it worked OK. What do I need to do to see the error that's being fixed here?

alyssawilk:
I think the issue explains better than I can - the problem is when you're trying to build android libraries on a mac machine it gets confused about platforms.  You can repro if you check out and build envoy mobile (or tensorflow but I can't help with that one) on a mac machine without the absl patch.  I'm happy to walk through repro instructions if you'd be game for trying it, or I can add a custom select so that we (and other folks who ran into the same issue) can work around the issue locally.

BUILD
devjgm: From reading the history of that other PR and issue that @devbww mentioned, I think the issue was that old versions of bazel didn't automatically link with `-framework Foundation`, and newer bazel's may do that automatically. I'm guessing that the bazel versions that we support today all implicitly link w/ Foundation, so I think removing these explicit deps may work OK.

That said, I'm not sure we actually _should_ do that. Today, our code explicitly references macOS TimeZone types, for example:

https://github.com/google/cctz/blob/c674330ee6c870ff0a77e965b1811de5514bb8ae/src/time_zone_lookup.cc#L25

and in the spirit of ~include~ link with what you use, cctz probably should be explicit about its dep on `Foundation`[1], IMO.

From https://github.com/abseil/abseil-cpp/issues/326, it sounds like the real issue is w/ the native android build rules. @alyssawilk did you say that you have an alternative workaround that may allow you to fix your issue while allowing CCTZ to retain its explicit dep?

[1]: We're currently linking with `Foundation`, but we actually only need to link with `CoreFoundation` (it's a lower-level library). So this dep should probably actually be `-framework CoreFoundation`... @alyssawilk does making that change fix your issue (probably not; but we might as well check)

alyssawilk: Our workaround is carrying an abseil.patch file* which removes the foundation files (which looks like what everyone else is doing).  I'm now trying to get rid of that patch file, either by removing the explicit dependency or making it optional because the code is switching repos and I'm trying to clean up all the backported patches that I can.  Worst case I can probably move the backported patch but I'd rather find a fix we're all Ok with, whether it be removing the dependency as outdated and known problematic, or having an opt-in bazel select. I'm happy to do whatever works best for you folks but I'd really prefer we fix abseil rather than expect everyone downstream to patch!

https://github.com/envoyproxy/envoy-mobile/blob/main/bazel/abseil.patch

devjgm: I agree that it would be good to avoid the need for everyone (anyone) to have to apply a patch in order to use abseil/cctz w/ an android build target (when compiling on macos).

I think I'd be OK with replacing the `-framework Foundation` explicit dep with a comment stating that we actually do need the `Foundation` dep, but it's added implicitly by Bazel and not listing it explicitly works around a bug in the android native deps (and link to that issue). That's basically this PR but with an additional comment.

@devbww WDYT?

@derekmauro can you confirm whether this PR, when patched in to Abseil, will pass all of Abseil's tests?

alyssawilk: SG thanks!  I'f we're doing that as general clean up I'm inclined to do it for both, not just for the dependency my project needs.  I'm very happy to revert that half if y'all prefer but I'll start with that so @derekmauro 's CI run tests both removals.

derekmauro: This passes CI: https://github.com/abseil/abseil-cpp/pull/1306

devjgm:
This PR LGTM. @devbww, thoughts?

devbww (commented):
> Perhaps the PR title could be more descriptive of the actual change.

Sorry for the confusion, but my concern was mostly with the title (first line), not the body.  It should try to summarize the actual effect.

BUILD
```diff
@@ -66,15 +52,9 @@ cc_library(
...
-            "-framework Foundation",
-        ],
-        "//:ios": [
-            "-framework Foundation",
-        ],
-        "//conditions:default": [],
-    }),
+    # Note that OS X and iOS both have an implicit dependency on Foundation but
+    # it is no longer explicitly added as a linkopt as bazel adds it
+    # automatically. See https://github.com/abseil/abseil-cpp/issues/326 for details.
```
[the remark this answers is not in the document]
alyssawilk: the way bazel adds it doesn't cause problems for android builds on macos.
Given this fixes known bugs for multiple downstream customers and the only down-side is an implicit rather than explicit dependency, would it be possible to file a tracking issue for you folks to investigate what bazel does and add back a (working) explicit dependency?  I don't know enough about bazel internals to know what they're doing and how it works but I know from again the multiple reports of breakage with explicit includes that there is a real problem I would like to fix, without the perfect solution stopping a known improvement.

BUILD
```diff
@@ -66,15 +52,9 @@ cc_library(
...
-    linkopts = select({
-        "//:osx": [
-            "-framework Foundation",
-        ],
-        "//:ios": [
-            "-framework Foundation",
-        ],
-        "//conditions:default": [],
-    }),
+    # Note that OS X and iOS both have an implicit dependency on Foundation but
```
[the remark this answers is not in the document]
alyssawilk: sorry I don't know what you mean by mention -framework CoreFoundation directly.  Could I get a bit more detail here?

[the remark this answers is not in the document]
alyssawilk: oh maybe thi sis just replacing Foundatoin with CoreFoundation?  I'll optimistically go for that :-)

BUILD
```diff
@@ -52,9 +52,10 @@ cc_library(
         "include/cctz/zone_info_source.h",
     ],
     includes = ["include"],
-    # Note that OS X and iOS both have a dependency on Foundation but it is no
-    # longer explicitly added as a linkopt as that causes problems cross-compiling android
-    # builds on OSX, and bazel will add it automatically.
+    # Note that OS X and iOS both have a dependency on CoreFoundation but it is
+    # no longer explicitly added as a linkopt as that causes problems
+    # cross-compiling android builds on OSX, and bazel will add it
+    # automatically.
```
devbww, line 58: OK, thanks.  If you have the stamina for one more edit, though, I'd suggest ...
```
    # OS X and iOS no longer use `linkopts = ["-framework CoreFoundation"]`
    # as (1) bazel adds it automatically, and (2) it caused problems when
    # cross-compiling for Android.
```

devbww (approved):
Thanks.

I also removed "attempting to" from the PR title.

BUILD
alyssawilk: SGTM!  I have no preference on phrasing so happy to land whatever works for you folks

sluongng:
Im a bit confused with this PR. I see many comments cited Bazel implicitly adding `-foundation` flag to the link opts but I don't see any reference to that.
When searching for it via https://cs.opensource.google/bazel, all I see is that Bazel "may" do it for objective-c compilation and not for cc.

My personal guess is that on MacOS, the default cc toolchain Bazel uses is the host's local cc toolchain (xcode clang). This toolchain is poorly configured and has leaks that may automatically include the framework where needed. However, when you switch to using a more hermetic, air-tight cc toolchain (such as https://github.com/uber/bazel-zig-cc), the missing of implicit `-foundation` flag is clear.

As suggested in https://github.com/envoyproxy/envoy/issues/24782, this PR might be a bit rushed and android issue might be unrelated/tangent to the linkopts flag. I would urge folks to reconsider and perhaps, revert this PR.

devbww:
@alyssawilk ... please comment.  Thanks.

alyssawilk:
Sorry would reverting the PR require everyone to maintain the workaround patches they used to?  I'm not understand what changed to make the prior behavior less problematic than it was.

bdhess:
I'm not sure what changed either, but there's definitely a problem here.  I have a real program running on macOS 13 / M1 that segfaults due to this issue.  We don't see this issue in Kokoro which uses Intel Macs.

```
Termination Reason:    Namespace DYLD, Code 4 Symbol missing
missing symbol called
(terminated at launch; ignore backtrace)

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   dyld                          	       0x184b69118 __abort_with_payload + 8
1   dyld                          	       0x184b74d7c abort_with_payload_wrapper_internal + 104
2   dyld                          	       0x184b74db0 abort_with_payload + 16
3   dyld                          	       0x184b008a8 dyld4::halt(char const*) + 328
4   dyld                          	       0x184b36770 dyld4::APIs::_dyld_missing_symbol_abort() + 44
5   aes_gcm_test                  	       0x100e4d618 absl::time_internal::cctz::local_time_zone() + 36
6   aes_gcm_test                  	       0x100e2e188 absl::LocalTimeZone() + 16
7   aes_gcm_test                  	       0x100e2e14c absl::FormatTime(absl::Time) + 56
8   aes_gcm_test                  	       0x100d8bc2c grpc_core::StatusToString(absl::Status const&)::$_0::operator()(std::__1::basic_string_view<char, std::__1::char_traits<char>>, absl::Cord const&) const + 1228
```

Is it possible that the "default Bazel behavior" wasn't a Bazel behavior at all, but actually some default in Xcode?

keith:
I submitted a revert for this here https://github.com/google/cctz/pull/264

PR #240, opened 2023-01-07
Fix build on cygwin

DarthGandalf:
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/240)


devbww (approved):
Thanks.

While we don't have CI for Cygwin (obviously), I have manually verified that this indeed fixes the link, with Cygwin now falling into the _'Make it a "strong" definition'_ case.

PR #241, opened 2023-02-27
Add an option to separate benchmark build from tests

sbraz:
Hi,
Here is a small PR that will help adding this package to Gentoo.
In some cases, it is desirable to run tests but no benchmark.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/cctz/241)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,14 +16,17 @@ include(FeatureSummary)
 
 option(BUILD_TOOLS "Whether or not to build tools" ON)
 option(BUILD_EXAMPLES "Whether or not to build examples" ON)
+option(BUILD_BENCHMARK "Whether or not to build the benchmark application (has no effect if BUILD_TESTING is disabled)" ON)
 
 if (BUILD_TESTING)
-  find_package(benchmark)
-  set_package_properties(benchmark PROPERTIES
-    TYPE REQUIRED
-    DESCRIPTION "a microbenchmark support library"
-    URL "https://github.com/google/benchmark"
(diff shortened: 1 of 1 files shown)


sbraz:
Opened https://github.com/google/oss-fuzz/pull/9811 to address the failing CI

devbww:
> Opened [google/oss-fuzz#9811](https://github.com/google/oss-fuzz/pull/9811) to address the failing CI

It seems like this may also be a problem for other users who currently use `BUILD_TESTING=OFF` to skip anything benchmark related.  Can we make it so that `BUILD_BENCHMARK` is a sub-condition of `BUILD_TESTING=ON`?

sbraz:
@devbww how about this? I'm not sure the long option description is a good idea but I thought it was worth mentioning.

devbww:
Seems reasonable.  I might only suggest using parentheses.  That is, _" (has no effect ...)"_ instead of _", has no effect ..."_.

I'll rerun the CI builds.

devbww:
@derekmauro ... any idea what's up with the kokoro builds.  They seem to be "Waiting" forever.

And sorry that this is probably the Nth time I've asked for your advice on such things.  :-(

derekmauro:
It seems like it may have just been a communication issue between GitHub and Kokoro. I applied the tag kokoro:force-run and it ran.

I sent you a change for the Linux CI script to use a newer Docker image. Once that it submitted we can re-run this.

devbww (approved):
Thanks.

PR #251, opened 2023-05-18
Fully qualify cctz::format in tests

derekmauro:
When cctz is imported into Abseil in a nested namespace (absl::time_internal::cctz), GCC 13 considers unqualified calls to cctz::format to be ambiguous with C++20's std::format.

I tried adding a using declaration at the top of the files, but that didn't work.


devbww (approved):
That's a shame.

PR #252, opened 2023-05-19
Update time_zone_lookup.cc

wony-zheng:
change read local timezone from window api, no need manual set enviroment variable

### DIFF
--- a/src/time_zone_lookup.cc
+++ b/src/time_zone_lookup.cc
@@ -33,6 +33,11 @@
 #include <zircon/types.h>
 #endif
 
+#if defined(_MSC_VER)
+#include <Windows.h>
+#include <timezoneapi.h>
+#endif
+
 #include <cstdlib>
 #include <cstring>
 #include <string>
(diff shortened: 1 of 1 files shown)


src/time_zone_lookup.cc
```diff
@@ -207,8 +212,9 @@ time_zone local_time_zone() {
   char* localtime_env = nullptr;
   if (strcmp(zone, "localtime") == 0) {
 #if defined(_MSC_VER)
-    // System-specific default is just "localtime".
-    _dupenv_s(&localtime_env, nullptr, "LOCALTIME");
+    TIME_ZONE_INFORMATION zi = {};
+    GetTimeZoneInformation(&zi);
+    return fixed_time_zone(seconds(-zi.Bias * 60));
```
devbww, line 217: I think this is largely antithetical to CCTZ's goal of providing accurate timezone information for a region.  Indeed, if I had my time again I would omit `fixed_time_zone()` from the API.  It is typically more a part of a problem rather than a solution.

src/time_zone_lookup.cc
```diff
@@ -207,8 +212,9 @@ time_zone local_time_zone() {
   char* localtime_env = nullptr;
   if (strcmp(zone, "localtime") == 0) {
 #if defined(_MSC_VER)
-    // System-specific default is just "localtime".
-    _dupenv_s(&localtime_env, nullptr, "LOCALTIME");
```
devbww, line 211: Using `${LOCALTIME}` to override the name _"localtime"_ is part of the API, and cannot be removed.

devbww (commented):
I believe you'll find this unnecessary after the recent merge of #242, where we now discover the IANA name of the configured Windows local timezone, and hence can use real tzdata without the need for additional settings like `${TZ}` or `${LOCALTIME}`.

PR #255, opened 2023-05-24
Fix sign-conversion warning

DanilChapovalov:
Add explicit static_cast UINT32 -> int and int -> size_t to support comiling cctz with -Wsign-conversion


devbww (approved):
Thanks.

PR #259, opened 2023-06-08
Add VxWorks support

wenshan1:
This PR adds VxWorks support. 
A little code is changed for:
1. strptime() is not supported by VxWorks.
2. The time zone variables in VxWorks are:  timezone and tzname.


wenshan1:
Close this PR, re-submit a new PR.

PR #260, opened 2023-06-08
Add VxWorks support

wenshan1:
This PR adds VxWorks support.

A little code is changed for:
strptime() is not supported by VxWorks.
The time zone variables in VxWorks are: timezone and tzname.


devbww (approved):
Thanks.

PR #264, opened 2023-07-17
Revert "build: fix cross-platform android builds on OS X (#237)"

keith:
This breaks toolchains that don't add the implicit Foundation linkopt. I can no longer reproduce this building envoy for Android so I'm not sure if we mis-diagnosed the issue, or if it was fixed another way.

https://github.com/google/cctz/pull/237

This reverts commit 5417654582642359a597b273bacaaedcc99f758f.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -52,10 +52,11 @@ cc_library(
         "include/cctz/zone_info_source.h",
     ],
     includes = ["include"],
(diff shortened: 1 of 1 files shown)


devbww (approved):
Thanks for following up.

PR #267, opened 2023-07-31
Extend transitions for a full 400 years

devbww:
2022g introduced a new zone, America/Ciudad_Juarez, whose "-b slim" form ends in a year (2022) with three transitions, and which therefore cannot fully match the POSIX TZ string used to handle instants after the last transition.

Previously we assumed that the last-year's transitions would match those generated by the POSIX string.  In this case that meant we claimed there was an extra transition in any future year equivalent to (2022 mod 400).

So, extend the transitions using the POSIX string for a full 400 years, instead of 399.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -363,7 +363,7 @@ bool TimeZoneInfo::ExtendTransitions() {
       if (last_time < ta->unix_time) transitions_.push_back(*ta);
       transitions_.push_back(*tb);
     }
-    if (last_year_ == limit) break;
+    if (last_year_ > limit) break;
     jan1_time += kSecsPerYear[leap_year];
     jan1_weekday = (jan1_weekday + kDaysPerYear[leap_year]) % 7;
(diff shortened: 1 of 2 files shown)


src/time_zone_info.cc
```diff
@@ -363,7 +363,7 @@ bool TimeZoneInfo::ExtendTransitions() {
       if (last_time < ta->unix_time) transitions_.push_back(*ta);
       transitions_.push_back(*tb);
     }
-    if (last_year_ == limit) break;
+    if (last_year_ > limit) break;
```
derekmauro, line 366: Just checking: should the reservation on line 341 be adjusted?

derekmauro (approved):

src/time_zone_info.cc
devbww: Yes.  Good catch.

And that led me to a better explanation and implementation, which I'll send in a new PR.

Thanks.

PR #269, opened 2023-08-04
Changes the Windows SDK version check to support MingW

derekmauro:
Windows SDK added `windows.globalization.h` file from SDK 10.0.10240.0, but MinGW added this file in 10.0.22621.0.

Credit to @veblush for co-developing this fix.

### DIFF
--- a/src/time_zone_lookup.cc
+++ b/src/time_zone_lookup.cc
@@ -37,7 +37,12 @@
 #include <sdkddkver.h>
 // Include only when the SDK is for Windows 10 (and later), and the binary is
 // targeted for Windows XP and later.
-#if defined(_WIN32_WINNT_WIN10) && (_WIN32_WINNT >= _WIN32_WINNT_WINXP)
+// Note: Windows SDK added windows.globalization.h file in 10.0.10240.0,
+//   but mingw added it in 10.0.22621.0, so the version is checked against
+//   10.0.22621.0 in mingw.
+#if ((defined(_WIN32_WINNT_WIN10) && !defined(__MINGW32__)) ||        \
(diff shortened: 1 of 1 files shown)


src/time_zone_lookup.cc
```diff
@@ -37,7 +37,12 @@
 #include <sdkddkver.h>
 // Include only when the SDK is for Windows 10 (and later), and the binary is
 // targeted for Windows XP and later.
-#if defined(_WIN32_WINNT_WIN10) && (_WIN32_WINNT >= _WIN32_WINNT_WINXP)
+// Note: Windows SDK added windows.globalization.h file in 10.0.10240.0,
+//   but mingw added it in 10.0.22621.0, so the version is checked against
+//   10.0.22621.0 in mingw.
```
devbww, line 42: None of these version numbers appear in the conditional, which makes it difficult to see what is being checked.  Should the note be phrased in terms of the symbols instead?

src/time_zone_lookup.cc
```diff
@@ -37,7 +37,12 @@
...
+#if ((defined(_WIN32_WINNT_WIN10) && !defined(__MINGW32__)) ||        \
+     (defined(NTDDI_WIN10_NI) && NTDDI_VERSION >= NTDDI_WIN10_NI)) && \
+    (_WIN32_WINNT >= _WIN32_WINNT_WINXP)
```
devbww, line 45: Should this repeated expression be encapsulated in a single, say, "USE_WIN32_LOCAL_TIME_ZONE"?

src/time_zone_lookup.cc
derekmauro: Done.

src/time_zone_lookup.cc
derekmauro: Done.

devbww (approved):

PR #278, opened 2023-10-11
[fuchsia] Adds new tzdata dir, explains all dirs

filmil:
Added `/config/tzdata` as a canonical path in the
Fuchsia component's namespace.

This is from:
https://fuchsia.dev/fuchsia-src/concepts/process/namespaces#typical_directory_structure

Each directory prefix entry in the list of prefixes that Fuchsia looks for timezone data files is now
explained with a comment.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -534,9 +534,16 @@ std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
   // Prefixes where a Fuchsia component might find zoneinfo files,
   // in descending order of preference.
   const auto kTzdataPrefixes = {
+      // The tzdata from `config-data`.
       "/config/data/tzdata/",
+      // The tzdata bundled in the component's package.
       "/pkg/data/tzdata/",
+      // General data storage.
       "/data/tzdata/",
+      // The recommended path for routed-in tzdata files.
+      // See for details:
+      // https://fuchsia.dev/fuchsia-src/concepts/process/namespaces?hl=en#typical_directory_structure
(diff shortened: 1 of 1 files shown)


src/time_zone_info.cc
```diff
@@ -534,9 +534,16 @@ std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
...
+      // The tzdata from `config-data`.
       "/config/data/tzdata/",
+      // The tzdata bundled in the component's package.
       "/pkg/data/tzdata/",
+      // General data storage.
       "/data/tzdata/",
+      // The recommended path for routed-in tzdata files.
+      // See for details:
+      // https://fuchsia.dev/fuchsia-src/concepts/process/namespaces?hl=en#typical_directory_structure
+      "/config/tzdata/",
```
devbww, line 546: Just checking that these are indeed in "descending order of preference".  Naively, it seems strange that "/config/data/tzdata/" and "/config/tzdata/" are on opposite ends of the spectrum.

filmil: In a way yes, this is the correct ordering. `/config/data/...` is deprecated, and `/data/tzdata` *may* be present in some older components, and `/pkg/data` should probably take precedence over `/config/tzdata` if the owners so wants.

devbww: From the way you describe things, it sounds like the ordering should be:

1. `/pkg/data/tzdata/`
2. `/config/tzdata/`
3. `/config/data/tzdata/`
4. `/data/tzdata`

But, it's your choice.

devbww (approved):

PR #285, opened 2024-02-20
Add support for DLLs on Windows

uilianries:
Greetings!

The CMake recommendation for RUNTIME destination is the `bin` folder, not lib. It occurs because it installs .dll files there and that folder will be integrated to the PATH environment.

https://cmake.org/cmake/help/latest/command/install.html#signatures

Plus, to export symbols to a DLL, CMake offers [CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) which does it automatic. So you do not need to add `__declspec` in your code with macros.

So, when passing `-DBUILD_SHARED_LIBS=ON`, the cctz, when build in Windows with msvc, it should result in `${CMAKE_INSTALL_PREFIX}/lib/cctz.lib` and `${CMAKE_INSTALL_PREFIX}/bin/cctz.dll`, where the .lib works like a glue to load the .dll


devbww (approved):
Thanks.

PR #291, opened 2024-05-24
RFC: Add number_to_weekday and weekday_to_number

teksturi:
Currently this is just suggestion. Example no unittests yet and some details
could be changed.

We have discussion in this PR [1] that weekday should not be casted. I check
our company codebase and we did cast weekday 9 places (oops). We definitely
need those cast as we are example converting them for external APIs. So for
this reason I implemented internally very similar functions as in this PR. I would
like to know if this is something could fit for cctz. Right place could also be
Abseil so please let me know what you think.

[1]: https://github.com/google/cctz/pull/290

### DIFF
--- a/include/cctz/civil_time_detail.h
+++ b/include/cctz/civil_time_detail.h
@@ -18,6 +18,7 @@
 #include <cstdint>
 #include <limits>
 #include <ostream>
+#include <string>
 #include <type_traits>
 
 // Disable constexpr support unless we are in C++14 mode.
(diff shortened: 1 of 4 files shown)


include/cctz/civil_time_detail.h
```diff
@@ -566,6 +567,198 @@ CONSTEXPR_F civil_day prev_weekday(civil_day cd, weekday wd) noexcept {
...
+
+enum class weekstart
+{
+  monday_zero,
+  monday_one,
+  sunday_zero,
+  sunday_one,
+};
+
+CONSTEXPR_F int weekday_to_number(weekday wd, weekstart ws) noexcept {
```
teksturi, line 580: weekstart could basically always be static. So I wonder if we can use template
so this would be

weekday_to_number<weekstart::xxx>();
or even
weekday_to_number<weekstart::xxx, int>();

but the way I now implemented it seems to be most aligned with cctz
current api.

include/cctz/civil_time_detail.h
```diff
@@ -566,6 +567,198 @@ CONSTEXPR_F civil_day prev_weekday(civil_day cd, weekday wd) noexcept {
...
+          return 7;
+      }
+      break;
+  }
+
+  // Undefined behavior.
+  return 0;
+}
+
+CONSTEXPR_F weekday number_to_weekday(int wd, weekstart ws, std::string* err = nullptr) noexcept {
```
teksturi, line 660: is std::string good fit? I tried to copy interface from parse(). Basically bool would
be enogh. I first also did that we return error and weekday is pointer but that did
not feel right as many times err is optional and programmer really knows already
that error cannot happend.

src/time_zone_format.cc
```diff
@@ -63,48 +63,6 @@ char* strptime(const char* s, const char* fmt, std::tm* tm) {
 }
 #endif
 
-// Convert a cctz::weekday to a tm_wday value (0-6, Sunday = 0).
-int ToTmWday(weekday wd) {
```
teksturi, line 67: I could also save these functions and just return weekday_to_number or number_to_weekday. Then we still have these names but that did not feel right as example this function has comment about that sunday = zero. When we use weekday_to_number straight we explicitly see that sunday is zero.

include/cctz/civil_time_detail.h
teksturi, line 580: Other possiblity could be to have

- weekday_to_sunday_zero
- weekday_to_sunday_one
- weekday_to_monday_zero
- weekday_to_monday_one

This use using there would be shorter and we would not need enum. Compare:
auto wd_num = cctz::weekday_to_number(1, cctz::weekstart::monday_zero);
auto wd_num = cctz::weekday_to_monday_zero(1);

derekmauro:
As you saw in https://github.com/google/cctz/pull/290#pullrequestreview-2027433314, we have consciously decided not to do this and do not intend to change this decision.

include/cctz/civil_time_detail.h
devbww: > We definitely need those cast as we are example converting them for external APIs.

> Other possiblity could be to have
> 
> * weekday_to_sunday_zero
> * weekday_to_sunday_one
> * weekday_to_monday_zero
> * weekday_to_monday_one

In addition to @derekmauro's comment, I would also suggest that such conversion functions belong in the converting code (where the "from" and "to" are clear, just like the existing `tm` case), rather than having N public helper functions in CCTZ.

teksturi:
Ok. I still see problem here. What about adding warning to weekday documentation that this should not be static casted. Else it will happened again and again.

derekmauro:
If you aren't familiar with [Hyrum's Law](https://www.hyrumslaw.com/), I suggest [learning about it](https://abseil.io/resources/swe-book/html/ch01.html#hyrumapostrophes_law).

The general rule for an `enum class` is to treat is strictly as a set of labels unless the contract says otherwise. `enum class` is not implicitly convertible to `int` and this is the motivation for this design choice.

If you need a mapping from `weekday` to `int`, I would suggest you should define that yourself. Different programs may require different mappings, there is no reason that we should define a mapping that we don't need and that may not work best for your application.

devbww:
My 2c:

I doubt anyone would heed a warning not to cast if they're set on doing so.  But we certainly have no plans to change the values, so at least whatever they do should continue to work.

> The general rule for an `enum class` is to treat is strictly as a set of labels unless the contract says otherwise. `enum class` is not implicitly convertible to `int` and this is the motivation for this design choice.

Exactly.  An `enum class` without explicit assignments, nor an explicit underlying type, should be a clear indication that the values  are immaterial.

> If you need a mapping from `weekday` to `int`, I would suggest you should define that yourself. Different programs may require different mappings, there is no reason that we should define a mapping that we don't need and that may not work best for your application.

Right.  And even so, I wouldn't suggest doing the mapping via a conversion to `int`.

I also doubt this will come up "again and again."  But whenever it does, I'll be happy to give the same advice.

Even though we declined this, thanks for the suggestion and for taking an interest in CCTZ.

PR #294, opened 2024-07-06
Use reference in range-for with non trivial type

gruenich:
Found by Clazy (range-loop-reference)


derekmauro (approved):

src/time_zone_info.cc
```diff
@@ -551,7 +551,7 @@ std::unique_ptr<ZoneInfoSource> FuchsiaZoneInfoSource::Open(
   const auto prefixes = name_absolute ? kEmptyPrefix : kTzdataPrefixes;
 
   // Fuchsia builds place zoneinfo files at "<prefix><format><name>".
-  for (const std::string prefix : prefixes) {
+  for (const std::string& prefix : prefixes) {
```
derekmauro, line 554: This change is actually incorrect. prefixes is an array of `const char[]`, so this binds a temporary.

PR #299, opened 2024-09-09
Add GitHub Actions CI

derekmauro:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,135 @@
...
+        --cxxopt=-Wvla
+        --cxxopt=-Wwrite-strings
+        --cxxopt=-Wno-missing-declarations
+        --cxxopt=-Wno-undef
+        --features=external_include_paths
+        --keep_going
+        --show_timestamps
+        --test_env="TZDIR=${GITHUB_WORKSPACE}/testdata/zoneinfo"
+        --test_output=errors
+        --test_tag_filters=-benchmark
```
dinord, line 46: If we have open source benchmarks, it would be good to at least build them.  
Do we still build them with this flag?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,135 @@
+name: ci
+
+on:
+  push:
+  pull_request:
+
+env:
+  BAZEL_CXXOPTS: -std=c++14
```
dinord, line 8: I'm assuming we'll add more language versions in the future--I'd leave a TODO here.

dinord (approved):
🎉🎉🎉🎉

.github/workflows/ci.yaml
derekmauro: Yes, we do build them.

.github/workflows/ci.yaml
derekmauro: This replicates the current build. I'll expand this in the future.

PR #301, opened 2024-09-20
Support for QNX 7.1

kheaactua:
Without specifying _XOPEN_SOURCE>=600, QNX 7.1 fails with:

```
/target/qnx7/usr/include/c++/v1/__threading_support: In function 'int std::__1::__libcpp_condvar_timedwait(std::__1::__libcpp_condvar_t*, std::__1::__libcpp_mutex_t*, timespec*)': /target/qnx7/usr/include/c++/v1/__threading_support:293:10: error: 'pthread_cond_timedwait' was not declared in this scope
   return pthread_cond_timedwait(__cv, __m, __ts);
          ^~~~~~~~~~~~~~~~~~~~~~
/target/qnx7/usr/include/c++/v1/__threading_support:293:10: note: suggested alternative: 'pthread_cond_wait'
   return pthread_cond_timedwait(__cv, __m, __ts);
          ^~~~~~~~~~~~~~~~~~~~~~
          pthread_cond_wait
```

### DIFF
--- a/src/time_zone_format.cc
+++ b/src/time_zone_format.cc
@@ -20,7 +20,11 @@
 
 #if defined(HAS_STRPTIME) && HAS_STRPTIME
 # if !defined(_XOPEN_SOURCE) && !defined(__FreeBSD__) && !defined(__OpenBSD__)
-#  define _XOPEN_SOURCE 500  // Exposes definitions for SUSv2 (UNIX 98).
+#  if defined(__QNX__)
+#    define _XOPEN_SOURCE 600  // Exposes definitions for QNX 7.1.
+#  else
+#    define _XOPEN_SOURCE 500  // Exposes definitions for SUSv2 (UNIX 98).
+#  endif
 # endif
 #endif
 


derekmauro (commented):
Can you post the full error message? The part of the error message that you posted mentions `pthread_cond_timedwait` but I don't see how that is used by this file. It also seems weird that if this change is to fix `pthread_cond_timedwait` that the `#define` is going into a block guarded by `#if defined(HAS_STRPTIME) && HAS_STRPTIME`.

kheaactua:
I'm reconsidering this PR.  My toolchain used to define `_QNX_SOURCE`, which would also enable these pthread functions.  I removed it from my toolchain based on the [docs](http://www.qnx.com/developers/docs/7.1/index.html#com.qnx.doc.neutrino.utilities/topic/q/qcc.html)

> There are additional variants of this option that enable vendor-specific compiler extensions: gnu99, gnu11, gnu++98, gnu++11, gnu++14, and gnu++17. If you use these variants, `_QNX_SOURCE` is defined automatically

As hard as I try, I can't show that this statement is in any way true.

Re-enabling this in a QNX toolchain (`add_compile_definitions(_QNX_SOURCE)`) solves this issue without any code change.

PR #305, opened 2024-12-31
Add single point of control for default copts and linkopts

jwnimmer-tri:
This is a follow-up from https://github.com/abseil/abseil-cpp/pull/1194#pullrequestreview-997035491.

This will allow absl to govern them in its vendored copy of cctz.

In absl, instead of copying this new `copts.bzl` file from cctz, we can make a de-novo `absl/time/internal/cctz/copts.bzl` file that loads the `ABSL...` constants and re-exports them to this spelling.

Of course I'm open to any other solutions, or if you prefer to write the new code yourself instead of this pull request.  My only aim is to propagate the options, one way or another.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -12,6 +12,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+load(":copts.bzl", "DEFAULT_COPTS", "DEFAULT_LINKOPTS", "TEST_COPTS")
+
 package(
     features = [
         "header_modules",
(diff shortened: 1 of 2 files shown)


derekmauro:
I think what I would like to do here instead is to fix the [Copybara](https://github.com/google/copybara) script that brings CCTZ into Abseil so that it adds `ABSL_DEFAULT_COPTS` and friends as expected.

jwnimmer-tri:
Sounds good, thanks.

Issue #53, opened 2017-10-06
`cctz::local_time_zone` returns UTC on windows


vspinu:
I have recently integrated CCTZ into R package [lubridate](https://github.com/tidyverse/lubridate/). So far everything worked as expected, except for the local timezone on windows.  [`cctz::local_time_zone()`](https://github.com/hadley/lubridate/blob/master/src/update.cpp#L29) returns UTC on that platform. 

R itself goes through quite [some pain](https://github.com/SurajGupta/r-source/blob/master/src/extra/tzone/registryTZ.c) to get the tzone on Windows. I wonder if something like that is within the scope of  CCTZ.

devjgm:
On Fri, Oct 6, 2017 at 9:10 AM, Vitalie Spinu <notifications@github.com>
wrote:

> I have recently integrated CCTZ into R package lubridate
> <https://github.com/tidyverse/lubridate/>. So far everything worked as
> expected, except for the local timezone on windows.
> cctz::local_time_zone()
> <https://github.com/hadley/lubridate/blob/master/src/update.cpp#L29>
> returns UTC on that platform.
>

UTC is the fallback if cctz can't find any other reasonable local time
zone. Today CCTZ looks for the local timezone by checking environment
variables. So setting TZ=Foo/Bar should make cctz::local_time_zone() work,
even on windows.

But now I see that Windows has a GetTimeZoneInformation()
<https://msdn.microsoft.com/en-us/library/windows/desktop/ms724421(v=vs.85).aspx>
function, and that'd probably be a better thing to use. The ideal solution
would be to also add a Windows-specific subclass of the cctz::TimeZoneIf
interface, which would look up the time-zone data in the Windows registry,
or wherever that data comes from on Windows. Coping the TimeZoneLibC
<https://github.com/google/cctz/blob/master/src/time_zone_libc.h> class
would be a good place to start. The reason we haven't done this already is
because we don't have any Windows dev machines nor the experience to hack
it together. We're hoping an experienced Windows developer might be
interested in contributing this support to CCTZ :-)


> R itself goes through quite some pain
> <https://github.com/SurajGupta/r-source/blob/master/src/extra/tzone/registryTZ.c>
> to get the tzone on Windows. I wonder if something like that is within the
> scope of CCTZ.
>
Yes. See comments above.

devbww:
Hi Vitalie,

I don't think we would want to try to maintain anything like TZtable[] from the referenced R code.

[...]

PR #316, opened 2025-05-01
Remove WinRT deps from `cctz::local_time_zone`

### LINKED ISSUE #315: Consider removing WinRT dependency on Windows
This is a proposal to reconsider the dependency on WinRT that was introduced in the following pull-request.

 * https://github.com/google/cctz/pull/242

Since the above commit was deployed we have seen both stability and performance issues in projects that directly or indirectly use cctz.

 * https://github.com/google/mozc/issues/856
 * https://github.com/abseil/abseil-cpp/issues/1760

Basically there is no guarantee that `RoInitialize` can be safely called when `cctz::local_time_zone` gets called. In theory cctz can have its own thread to safely call `RoInitialize` and WinRT APIs, but such an approach would make the code much more complex and fragile.

My proposal is to use `ucal_getDefaultTimeZone` instead, as @mwinterb has pointed out in https://github.com/google/cctz/pull/242#issuecomment-1453857694
> I don't have much skin in the Windows C++ and timezones game anymore, but even more recent versions of Windows 10 come with ICU, and it seems like ucal_getDefaultTimeZone is a lot simpler. You'll need to LoadLibrary and GetProcAddress and all that nonsense, and that may include falling back to still needing to CoInitialize for older recent versions of Windows 10, but at least you don't need to track as many lifetimes.
> See the "Important bonus chatter" section here on versions here:
> https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255
>
> I can write up a "counter" proposal, but if you do need to support Server 2016 or older versions of Windows 10, then it doesn't matter if ucal_getDefaultTimeZone is simpler :)

yukawa:
This reworks our previous commit (1bb996caf91effbf3e74ac444b0945b6360a8822), which aimed to address #53 by implementing `cctz::local_time_zone` with WinRT API on Windows.

The issue is there is no guarantee that `RoInitialize` can be safely called when `cctz::local_time_zone` gets called. In theory cctz can have its own thread to safely call `RoInitialize` and WinRT APIs, but such an approach would make the code much more complex and fragile.

This commit takes advantage of newly exposed ICU APIs that do not require COM/WinRT initialization on Windows 10 1903 and later.

 * https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255

To maintain binary compatibility with earlier versions of Windows, the DLL and function pointer are still loaded dynamically.

Closes #315.

### DIFF
--- a/src/time_zone_lookup.cc
+++ b/src/time_zone_lookup.cc
@@ -31,23 +31,23 @@
 #endif
 
 #if defined(_WIN32)
-#include <sdkddkver.h>
-// Include only when the SDK is for Windows 10 (and later), and the binary is
-// targeted for Windows XP and later.
-// Note: The Windows SDK added windows.globalization.h file for Windows 10, but
-// MinGW did not add it until NTDDI_WIN10_NI (SDK version 10.0.22621.0).
-#if ((defined(_WIN32_WINNT_WIN10) && !defined(__MINGW32__)) ||        \
-     (defined(NTDDI_WIN10_NI) && NTDDI_VERSION >= NTDDI_WIN10_NI)) && \
-    (_WIN32_WINNT >= _WIN32_WINNT_WINXP)
+// Include only when <icu.h> is available.
+// https://learn.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-
+// https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255
+#if defined(__has_include)
+#if __has_include(<icu.h>)
 #define USE_WIN32_LOCAL_TIME_ZONE
-#include <roapi.h>
-#include <tchar.h>
-#include <wchar.h>
-#include <windows.globalization.h>
+#pragma push_macro("NTDDI_VERSION")
+// ucal_getTimeZoneIDForWindowsID is available on Windows 10 RS3 and later.
+#define NTDDI_VERSION 0x0A000004  // == NTDDI_WIN10_RS3
 #include <windows.h>
-#include <winstring.h>
-#endif
-#endif
+#include <icu.h>
+#include <timezoneapi.h>
+#pragma pop_macro("NTDDI_VERSION")
+#include <atomic>
+#endif  // __has_include(<icu.h>)
+#endif  // __has_include
+#endif  // _WIN32
 
 #include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


src/time_zone_lookup.cc
```diff
@@ -60,80 +60,78 @@ namespace cctz {
...
-  time_zone->Release();
-  calendar->Release();
-  return result;
+
+  const int num_bytes_in_utf8 = ::WideCharToMultiByte(
+      CP_UTF8, 0, reinterpret_cast<const wchar_t*>(buffer),
+      static_cast<int>(num_chars_in_buffer), nullptr, 0, nullptr, nullptr);
+  std::string local_time_str;
+  local_time_str.resize(static_cast<size_t>(num_bytes_in_utf8));
+  ::WideCharToMultiByte(CP_UTF8, 0, reinterpret_cast<const wchar_t*>(buffer),
```
derekmauro, line 128: https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte

Maybe this is somewhat pedantic because it will still work, but it seems like this writes the NUL character to [local_time_str.size()-1]. Which I would think would mean you should erase the last char to make local_time_str.size() consistent with the real size.

src/time_zone_lookup.cc
```diff
@@ -60,80 +60,78 @@ namespace cctz {
...
-  if (!windows_delete_string) {
-    return result;
-  }
-  const auto windows_get_string_raw_buffer =
-      reinterpret_cast<decltype(&WindowsGetStringRawBuffer)>(
-          GetProcAddress(combase, "WindowsGetStringRawBuffer"));
-  if (!windows_get_string_raw_buffer) {
-    return result;
+// True if we have already failed to load the API.
+static std::atomic_bool g_ucal_getTimeZoneIDForWindowsIDUnavailable;
```
derekmauro, line 64: Is it possible to factor the initialization code into some function called once to avoid these atomics?

src/time_zone_lookup.cc
```diff
@@ -60,80 +60,78 @@ namespace cctz {
...
-                          &result[0], size, nullptr, nullptr);
-    }
-    windows_delete_string(tz_hstr);
+  UChar buffer[128];
+  UErrorCode status = U_ZERO_ERROR;
+  const auto num_chars_in_buffer = ucal_getTimeZoneIDForWindowsIDFunc(
+      reinterpret_cast<const UChar*>(info.TimeZoneKeyName), -1, nullptr,
+      buffer, ARRAYSIZE(buffer), &status);
+  if (status != U_ZERO_ERROR || num_chars_in_buffer <= 0 ||
+      num_chars_in_buffer > ARRAYSIZE(buffer)) {
```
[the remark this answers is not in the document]
yukawa: I locally tested several buffer size for the same local time (`"Asia/Tokyo"`) and here is what I observed in my Windows 11 24H2 environment.

| `ARRAYSIZE(buffer)` | `num_chars_in_buffer` | `status` | content of `buffer`|
|----------|--------|-----------|--------|
|12 | 10 | `U_ZERO_ERROR` | fully written **with** NUL termination|
|11 | 10 | `U_ZERO_ERROR` | fully written **with** NUL termination|
|10 | 10 | `U_STRING_NOT_TERMINATED_WARNING` | fully written **without** NUL termination|
|9 | 10 |  `U_BUFFER_OVERFLOW_ERROR` | nothing is written|

Note that the subsequent code expect `num_chars_in_buffer` to not include NUL, so I think `num_chars_in_buffer > ARRAYSIZE(buffer)` makes sense in this case. `status != U_ZERO_ERROR` check effectively covers such cases already though.

Do you have any preference on this condition? Realistically `if (status != U_ZERO_ERROR)` seems to be sufficient. Happy to update the patch if you have any preference.

src/time_zone_lookup.cc
yukawa: In this particular case, `num_chars_in_buffer` is the UTF-16 char count without NUL so it falls into the following rule.

https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte
> If this parameter is set to a positive integer, the function processes exactly the specified number of characters. If the provided size does not include a terminating null character, the resulting character string is not null-terminated, and the returned length does not include this character.

So I think this is working as expected, right?

src/time_zone_lookup.cc
yukawa: > Is it possible to factor the initialization code into some function called once to avoid these atomics?

I guess here is what you were wondering about.

```cpp
std::string win32_local_time_zone() {
  // To avoid frequent DLL loadings and unloading, we obtain the API only once
  // then allow "icu.dll" to be "leaked". Given that it is a system library,
  // keeping it on memory is supposed to have no major drawback.
  const static auto ucal_getTimeZoneIDForWindowsIDFunc =
      []() -> decltype(&::ucal_getTimeZoneIDForWindowsID) {
    const HMODULE icudll =
        ::LoadLibraryExW(L"icu.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
    if (icudll == nullptr) {
      return nullptr;
    }
    return reinterpret_cast<decltype(&::ucal_getTimeZoneIDForWindowsID)>(
        ::GetProcAddress(icudll, "ucal_getTimeZoneIDForWindowsID"));
  }();

  DYNAMIC_TIME_ZONE_INFORMATION info = {};
  if (::GetDynamicTimeZoneInformation(&info) == TIME_ZONE_ID_INVALID) {
    return "";
  }

  ...
```

One concern in this approach is that thread-safe local static initialization does internally use a lock, which may cause dead lock when combined with Win32 Loader lock.

 * https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices#deadlocks-caused-by-lock-order-inversion

Suppose there are two running thread X and thread Y.

 1. Thread X enters `DllMain`, where the operating system implicitly acquires Loader lock.
 2. Thread Y enters `win32_local_time_zone()` then acquires the lock to initialize `ucal_getTimeZoneIDForWindowsIDFunc` variable. It then get blocked at `LoadLibraryExW` because the thread X still holds the Loader Lock.
 3. Thread X also enters `win32_local_time_zone()` then gets blocked because `ucal_getTimeZoneIDForWindowsIDFunc` is not yet fully initialized.

The current approach with `std::atomic_*` doesn't have the above dead lock issue, as it is just an optimistic optimization that does not try to enforce the initialization happens exactly once.

Let me know if you still prefer the thread-safe local static initialization approach. I'll update the patch accordingly.

derekmauro (approved):
You've addressed all of my concerns. Thanks.

derekmauro:
When I try to import this into Abseil, I get
```
ERROR: T:/src/git/abseil-cpp/absl/time/internal/cctz/BUILD.bazel:40:11: Compiling absl/time/internal/cctz/src/time_zone_lookup.cc failed: (Exit 2): cl.exe failed: error executing CppCompile command (from target //absl/time/internal/cctz:time_zone) C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\bin\HostX64\x64\cl.exe ... (remaining 1 argument skipped)
C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\um\winnt.h(169): fatal error C1189: #error:  "No Target Architecture"
```

yukawa:
Another thing that may work is to define `_WIN32_WINNT` as well as `NTDDI_VERSION`.

```cpp
#if defined(_WIN32)
// Include only when <icu.h> is available.
// https://learn.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-
// https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255
#if defined(__has_include)
#if __has_include(<icu.h>)
#define USE_WIN32_LOCAL_TIME_ZONE
#pragma push_macro("_WIN32_WINNT")
#pragma push_macro("NTDDI_VERSION")
// ucal_getTimeZoneIDForWindowsID is available on Windows 10 RS3 and later.
#define _WIN32_WINNT 0x0A00  // == _WIN32_WINNT_WIN10
#define NTDDI_VERSION 0x0A000004  // == NTDDI_WIN10_RS3
#include <windows.h>
#include <icu.h>
#include <timezoneapi.h>
#pragma pop_macro("NTDDI_VERSION")
#pragma pop_macro("_WIN32_WINNT")
#include <atomic>
#endif  // __has_include(<icu.h>)
#endif  // __has_include
#endif  // _WIN32
```

yukawa:
@derekmauro Can you check if the following change works for abseil?

 * https://github.com/google/cctz/pull/317

derekmauro:
It appears to work. Thanks!

PR #319, opened 2025-05-15
Disable fsanitize=array-bounds for days_per_month

fmayer:
This works around a missed optimization in clang, which does not see that m (when the whole chain is inlined into BreakTime) is always inbounds.


fmayer:
@derekmauro PTAL

include/cctz/civil_time_detail.h
```diff
@@ -91,6 +91,15 @@ CONSTEXPR_F int days_per_4years(int yi) noexcept {
 CONSTEXPR_F int days_per_year(year_t y, month_t m) noexcept {
   return is_leap_year(y + (m > 2)) ? 366 : 365;
 }
+// The compiler cannot optimize away the check if we use
+// fsanitize=array-bounds.
+// When this function is inlined, m is guaranteed to be in [1, 13).
```
devbww, line 96: Above we say `using month_t = std::int_fast8_t;   // [1:12]`, so we might as well use the same syntax, _"[1:12]"_, for consistency's sake.

devbww, line 96:
```
// When this function is inlined, ...
```
Is inlining germane to the issue?

include/cctz/civil_time_detail.h
```diff
@@ -91,6 +91,15 @@ CONSTEXPR_F int days_per_4years(int yi) noexcept {
...
+// To reduce the overhead, we use no_sanitize to skip the  unnecessary
```
devbww, line 97: `s/<sp><sp>/<sp>/`

devbww (commented):
Just some nits for my part.

include/cctz/civil_time_detail.h
fmayer: yes. just looking at the function without inlining, it is clear that we don't know anything about `m`. but even when we inline, and the compiler should know be able to optimize, it doesn't.

fmayer: reworded the comment.

fmayer: done

emomaxd (approved):
LGTM

---

**_include/cctz/civil_time_detail.h_**

```cpp
// Type aliases that indicate normalized argument values.
using month_t = std::int_fast8_t;   // [1:12]
```

derekmauro (commented):
I don't see anything wrong with doing this, but I do have to ask why? Is this function really that hot?

derekmauro:
> Can you expand on this? Maybe share some numbers and tell us how you measured it?

@fmayer privately shared the data with me.

emomaxd (approved):

derekmauro (approved):

PR #320, opened 2025-05-20
Fix GCC attribute warning introduced by #319

derekmauro:
`error: 'array-bounds' attribute directive ignored [-Werror=attributes]`


derekmauro:
@fmayer

fmayer:
Thanks. Sorry about that.

PR #326, opened 2025-08-04
Adds a time zone fuzz test

derekmauro:
Based on
https://github.com/google/oss-fuzz/blob/master/projects/cctz/fuzz_cctz.cc
but uses the Google FuzzTest framework and runs directly in the CCTZ project using Bazel


src/time_zone_test_util.h
dinord, line 1: I'd just call this library `time_zone_names.{h,cc}`  to avoid putting more symbols in it over time. It's already pretty large, and hard to look at.

dinord (approved):

src/time_zone_test_util.h
derekmauro: I went with `test_time_zone_names.{h,cc}` since we have a file with the same name in the internal repository.

PR #338, opened 2026-02-19
Fix cctz::format()/parse() to treat NULs like any other chars

devbww:
The previous commit (PR #337) was couched in terms of `parse()` rejecting input strings containing NULs.  Instead, we should treat NULs like other characters by allowing them in both the format and input strings.

Similarly, `format()` should allow NULs in its format string.

Also:
 - Simplify `FormatTM()` calls by handling an empty format string internally, rather than by guarding each invocation.
 - Utilize the `std::string` iterator-pair constructor and `append()`.

### DIFF
--- a/src/time_zone_format.cc
+++ b/src/time_zone_format.cc
@@ -40,7 +40,6 @@
 #include <ctime>
 #include <limits>
 #include <string>
-#include <vector>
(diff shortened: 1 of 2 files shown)


derekmauro (approved):
Thank you. This is a more technically principled approach.

PR #339, opened 2026-03-02
Update zoneinfo files to 2026a

devbww:
Moldova (Europe/Chisinau, and its link Europe/Tiraspol) has actually used EU transition times since 2022, so it shifts into and out of daylight time at 03:00 and 04:00 respectively, not 02:00 and 03:00.

zic no longer generates a no-op transition when simultaneous Rule and Zone changes cancel each other out, which affects Asia/Tbilisi.


derekmauro (approved):
Thanks again!

PR #340, opened 2026-03-15
Fix inverted null check in `time_zone_name_win.cc`

yukawa:
This commit follows up the previous commit (27ca17317e9e5772bcf30d5e21cf837b5525442a), which aimed to improve the code but ended up introducing an inverted null check in time_zone_name_win.cc. As a result, `LoadIcuGetTimeZoneIDForWindowsID()` currently returns `nullptr` when it should be returning a valid time zone ID.

This is also a preparation for implementing TimeZoneIf with Windows time APIs (#328).


src/time_zone_name_win.cc
```diff
@@ -112,7 +112,7 @@ ucal_getTimeZoneIDForWindowsID_func LoadIcuGetTimeZoneIDForWindowsID() {
   const auto ucal_getTimeZoneIDForWindowsIDRef =
       AsProcAddress<ucal_getTimeZoneIDForWindowsID_func>(
           icu_dll, "ucal_getTimeZoneIDForWindowsID");
-  if (ucal_getTimeZoneIDForWindowsIDRef != nullptr) {
+  if (ucal_getTimeZoneIDForWindowsIDRef == nullptr) {
```
devbww, line 115: I'm fine with the change, of course, but it does seem to indicate that we don't have anything that tests whether `GetWindowsLocalTimeZone()` works.

yukawa: Right. The existing fallback chain in `local_time_zone()` makes it difficult for the test to actually fail. I've added a dedicated unit test in cbbfd2808fbab48114c99f173cdef025e3c378d2. Does it make sense?

CMakeLists.txt
```diff
@@ -148,12 +148,26 @@ if (BUILD_TESTING)
...
+      ${CMAKE_THREAD_LIBS_INIT}
+      )
+    add_test(time_zone_name_win_test time_zone_name_win_test)
+  endif()
+
   # tests runs on testdata
+  set(CCTZ_TESTS civil_time_test time_zone_format_test time_zone_lookup_test)
+  if(WIN32)
+    list(APPEND CCTZ_TESTS time_zone_name_win_test)
+  endif()
```
devbww: Perhaps separate `list(APPEND CCTZ_TESTS ...)` calls would be best factored out to after the corresponding `add_test()` calls?  That would make each block its own separate unit, and remove the need for the extra `if(WIN32)` conditional.

yukawa: Sounds good. Done in 841a08a83ad886cc00450eeb31b58b758e4edeea.

src/time_zone_name_win_test.cc
```diff
@@ -0,0 +1,43 @@
...
+
+TEST(TimeZoneNameWin, GetWindowsLocalTimeZone) {
+  // On Windows 10 1809+ (where icu.dll is available in System32),
+  // GetWindowsLocalTimeZone() should return a valid IANA time zone name.
+  // Note that LOAD_LIBRARY_SEARCH_SYSTEM32 is not sufficient to reliably load
+  // "icu.dll" in the production code, but it should be OK for testing purposes.
+  HMODULE icu_dll =
+      ::LoadLibraryExW(L"icu.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
+  if (icu_dll != nullptr) {
+    const std::string tz = GetWindowsLocalTimeZone();
```
devbww: Perhaps factor this repeated statement out of the conditional.

yukawa: Done in d09a3fd29fe67fb49160975a947300b21f0961f1.

src/time_zone_name_win_test.cc
```diff
@@ -0,0 +1,42 @@
...
+  // "icu.dll" in the production code, but it should be OK for testing purposes.
+  HMODULE icu_dll =
+      ::LoadLibraryExW(L"icu.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
+  const std::string tz = GetWindowsLocalTimeZone();
+  if (icu_dll != nullptr) {
+    EXPECT_FALSE(tz.empty());
+    ::FreeLibrary(icu_dll);
+  } else {
+    EXPECT_TRUE(tz.empty());
+  }
```
devbww, line 39: My final question before handing off to @derekmauro.

Do we know which branch our Windows CI actually takes?  That is, is it checking that `GetWindowsLocalTimeZone()` succeeds or fails?

src/time_zone_name_win.cc
yukawa: Marking as resolved.

CMakeLists.txt
yukawa: Marking as resolved.

src/time_zone_name_win_test.cc
yukawa: Marking as resolved.

src/time_zone_name_win_test.cc
yukawa: It should go to `icu_dll != nullptr` route, so `GetWindowsLocalTimeZone()` should now be continuously tested in CI.

According to a [recent run](https://github.com/google/cctz/actions/runs/24900316780/job/72916003007), runner image is `windows-2025`, which should have `icu.dll` there.
> Runner Image
>   Image: windows-2025

yukawa: I'm going to mark this thread as resolved. Feel free to reopen if you have any remaining concern.

yukawa:
@devbww @derekmauro : Let me know if you have any remaining question on this PR.

derekmauro (approved):

PR #341, opened 2026-04-24
Update zoneinfo files to 2026b

derekmauro:
British Columbia (America/Vancouver) moved to permanent -07 on 2026-03-09

https://lists.iana.org/hyperkitty/list/tz-announce@iana.org/thread/VX2Z3CBO6KHTYZNBBKFFWM7ZCI6TVCXP/

PR #342, opened 2026-04-24
Update zoneinfo files to 2026b

devbww:
British Columbia (America/Vancouver, and its link Canada/Pacific) moved to permanent -07 on 2026-03-09 (although the source change was temporarily set to occur on 2026-11-01 instead, when it first makes a difference).

zic now identifies when one zone abbreviation is a suffix of another, allowing them to share table space.  This saves 4 bytes in Asia/Ho_Chi_Minh (and its link Asia/Saigon), Asia/Phnom_Penh, and Asia/Vientiane.


derekmauro:
Thanks. I just merged #341.

PR #346, opened 2026-07-10
Enforce TZ_MAX_TYPES limit and use size_t for iteration in TimeZoneInfo::Load() to avoid wrap-around at 256

higher-performance:
Otherwise there is a potential denial-of-service vulnerability using a crafted zoneinfo file that has 256 or more transition types that are all marked as DST.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -50,6 +50,7 @@
 #include "cctz/civil_time.h"
 #include "time_zone_fixed.h"
 #include "time_zone_posix.h"
+#include "tzfile.h"
 
 namespace cctz {
 
@@ -150,6 +151,7 @@ bool Header::Build(const tzhead& tzh) {
   timecnt = static_cast<std::size_t>(v);
   if ((v = Decode32(tzh.tzh_typecnt)) < 0) return false;
   typecnt = static_cast<std::size_t>(v);
+  if (typecnt > TZ_MAX_TYPES) return false;
   if ((v = Decode32(tzh.tzh_charcnt)) < 0) return false;
   charcnt = static_cast<std::size_t>(v);
   if ((v = Decode32(tzh.tzh_leapcnt)) < 0) return false;
@@ -722,7 +724,7 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
   // Determine the before-first-transition type.
   default_transition_type_ = 0;
   if (seen_type_0 && hdr.timecnt != 0) {
-    std::uint_fast8_t index = 0;
+    std::size_t index = 0;
     if (transition_types_[0].is_dst) {
       index = transitions_[0].type_index;
       while (index != 0 && transition_types_[index].is_dst)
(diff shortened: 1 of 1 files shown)


mkruskal-google (approved):

src/time_zone_info.cc
mkruskal-google: clang warnings from here

src/time_zone_info.cc
```diff
@@ -150,6 +151,7 @@ bool Header::Build(const tzhead& tzh) {
   timecnt = static_cast<std::size_t>(v);
   if ((v = Decode32(tzh.tzh_typecnt)) < 0) return false;
   typecnt = static_cast<std::size_t>(v);
+  if (typecnt > TZ_MAX_TYPES) return false;
```
mkruskal-google, line 154: Is this reachable in a test?

higher-performance: I don't think so, but I didn't see a clean way to test this. IIUC I'd have to pollute the filesystem to make the test load garbage to trigger this. Any thoughts?

src/time_zone_info.cc
higher-performance: Thanks, added a cast.

src/time_zone_info.cc
mkruskal-google: Ah yea ok, I finally traced out where these files are coming from and see the issue.  We *could* create a test that overrides zone_info_source_factory with a strong symbol to inject test data instead of reading from the filesystem.  It might be useful to have that in place to test more stuff in here

higher-performance: Oh I didn't even know we had that, I hadn't dug that far. If you have an idea how to do it, do you feel like taking a stab at setting that up in another PR or something so I can just base this on top?

mkruskal-google (approved):

src/time_zone_info.cc
mkruskal-google: Yea I think we can just follow up on that separately, for https://github.com/google/cctz/pull/345 too

higher-performance: Okay thanks. Should I wait for that or should I merge this? (just saw you approved)

mkruskal-google (approved):

src/time_zone_info.cc
mkruskal-google: Yea just merge it, no reason to hold back a vuln fix due to missing test infra

higher-performance: Yeah sounds good, thanks.

higher-performance:
Actually it seems I might not have merge permissions at all. Do you want to merge, or give me permissions?

src/time_zone_info.cc
```diff
@@ -722,7 +724,7 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
   // Determine the before-first-transition type.
   default_transition_type_ = 0;
   if (seen_type_0 && hdr.timecnt != 0) {
-    std::uint_fast8_t index = 0;
+    size_t index = 0;
```
devbww: Throughout this project (except in a few cases that slipped by me), we prefer `std::size_t`.

devbww (changes requested):

src/time_zone_info.cc
higher-performance: Done.

src/time_zone_info.cc
devbww, line 154: This is now redundant with lines 675/676, and should be removed.

Alternatively, we could move the `typecnt` and `charcnt` checks here too.

higher-performance: Oh wait I looked at the wrong lines, sorry. Just saw the same checks below.

higher-performance: Opened #348.

PR #348, opened 2026-07-11
Remove redundant check

higher-performance:
Added in #346


mkruskal-google:
Whoops, race condition with Derek's change.  I should've caught that on review >.<

mkruskal-google (approved):

PR #350, opened 2026-07-15
pass unsigned char to ctype functions in parse()

rajath201:
Repro: call `cctz::parse("%Y-%m-%d", "\xA0...", ...)` with any untrusted input byte >= 0x80.
Cause: the leading-whitespace skip and the specifier scanners hand `*data`/`*fmt` (a plain `char`) straight to `std::isspace`/`std::isdigit`. Where `char` is signed those bytes become negative ints, which is outside the ctype domain (unsigned char or EOF), so the call is UB — a debug MSVC CRT asserts and some libc builds index their classification table out of bounds.
Fix: cast each ctype argument to `unsigned char` at every call site in `parse()`/`format()` and the same construct in `time_tool.cc`. Valid input is unaffected; a leading 0xA0 is now consistently not treated as whitespace. Added a parse regression test covering non-ASCII input.

### DIFF
--- a/src/time_tool.cc
+++ b/src/time_tool.cc
@@ -239,9 +239,11 @@ const char* Basename(const char* p) {
 
 // std::regex doesn't work before gcc 4.9.
 bool LooksLikeNegOffset(const char* s) {
-  if (s[0] == '-' && std::isdigit(s[1]) && std::isdigit(s[2])) {
+  if (s[0] == '-' && std::isdigit(static_cast<unsigned char>(s[1])) &&
+      std::isdigit(static_cast<unsigned char>(s[2]))) {
     int i = (s[3] == ':') ? 4 : 3;
-    if (std::isdigit(s[i]) && std::isdigit(s[i + 1])) {
+    if (std::isdigit(static_cast<unsigned char>(s[i])) &&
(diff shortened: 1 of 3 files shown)


devbww (changes requested):
Looking good.  Thanks.

But let's add `isdigit()`/`isspace()` helpers to `src/time_zone_format()` (in the anonymous namespace), and an `isdigit()` helper to `src/time_tool.cc`, to reduce the casting clutter and help avoid similar problems in the future.

```
bool isdigit(char ch) {
  return std::isdigit(static_cast<unsigned char>(ch)) != 0;
}

bool isspace(char ch) {
  return std::isspace(static_cast<unsigned char>(ch)) != 0;
}
```

rajath201:
Makes sense. Added the isdigit()/isspace() helpers to the anonymous namespace in time_zone_format.cc and an isdigit() to time_tool.cc, and switched all the call sites over to them. The format test suite still passes (42/42) and time_tool behaves the same.

derekmauro:
FYI for this and your other PRs, we need you to sign the CLA before we can accept this.  https://cla.developers.google.com/

devbww (approved):
Thanks.

derekmauro (approved):

PR #351, opened 2026-07-16
Expand the test cases for ParsePosixSpec()

devbww:
Now that #347 has introduced `time_zone_posix_test`, expand it to cover all of `ParsePosixSpec()`, including success and failure cases.

Also:
 - Update the BNF for `ParsePosixSpec()` after #347 moved the comma into `datetime`.
 - Update the Open Group URL for the TZ environment-variable specification.


derekmauro (approved):
LGTM, Thanks!

PR #353, opened 2026-07-17
Prevent load_time_zone() path traversal or arbitrary fopen

derekmauro:
FileZoneInfoSource::Open() strips an optional file: prefix and then concatenates the remainder of the zone-name string under TZDIR (or uses it verbatim if absolute) with no canonicalisation or ".." rejection, then calls fopen(path, "rb"). An attacker-controlled zone name like "../../../../../../tmp/evil" escapes TZDIR. "file:/dev/stdin" or any FIFO blocks the calling thread indefinitely.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -385,11 +385,11 @@ namespace {
 
 using FilePtr = std::unique_ptr<FILE, int(*)(FILE*)>;
 
-// fopen(3) adaptor.
-inline FilePtr FOpen(const char* path, const char* mode) {
+// fopen(3) adaptor for reading zoneinfo files (read-only binary mode).
+inline FilePtr FOpen(const char* path) {
 #if defined(_MSC_VER)
   FILE* fp;
-  if (fopen_s(&fp, path, mode) != 0) fp = nullptr;
+  if (fopen_s(&fp, path, "rb") != 0) fp = nullptr;
   return FilePtr(fp, fclose);
 #else
   // Open non-blocking and verify the target is a regular file before handing it
@@ -397,20 +397,32 @@ inline FilePtr FOpen(const char* path, const char* mode) {
   // fopen() on a FIFO or device node (reachable via the "file:" prefix or an
   // absolute path) would block indefinitely or read unbounded data.
   const int fd = open(path, O_RDONLY | O_NONBLOCK | O_CLOEXEC);
-  if (fd < 0) {
-    return FilePtr(nullptr, fclose);
-  }
-  struct stat st;
-  if (fstat(fd, &st) != 0 || !S_ISREG(st.st_mode)) {
-    close(fd);
-    return FilePtr(nullptr, fclose);
-  }
-  FILE* fp = fdopen(fd, mode);
-  if (fp == nullptr) {
+  if (fd >= 0) {
+    struct stat st;
+    if (fstat(fd, &st) == 0 && S_ISREG(st.st_mode)) {
+      FILE* fp = fdopen(fd, "rb");
+      if (fp != nullptr) return FilePtr(fp, fclose);
+    }
     close(fd);
   }
-  return FilePtr(fp, fclose);
+  return FilePtr(nullptr, fclose);
 #endif
+}
+
+// Returns true if the zone name contains a ".." path-traversal component.
+inline bool HasPathTraversal(const std::string& name, std::size_t pos) {
+  // Exact match: ".."
+  if (name.compare(pos, std::string::npos, "..") == 0) return true;
+  // Leading component: "../"
+  if (name.compare(pos, 3, "../") == 0) return true;
+  // Interior component: "/../"
+  if (name.find("/../", pos) != std::string::npos) return true;
+  // Trailing component: "/.."
+  if (name.size() >= pos + 3 &&
+      name.compare(name.size() - 3, 3, "/..") == 0) {
+    return true;
+  }
+  return false;
 }
 
 // A stdio(3)-backed implementation of ZoneInfoSource.
@@ -452,7 +464,7 @@ std::unique_ptr<ZoneInfoSource> FileZoneInfoSource::Open(
 
   // Reject path-traversal components so that a relative zone name cannot
   // escape the zoneinfo directory (e.g., "../../etc/passwd").
-  if (name.find("..", pos) != std::string::npos) {
+  if (HasPathTraversal(name, pos)) {
     return nullptr;
   }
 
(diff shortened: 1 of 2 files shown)


src/time_zone_info.cc
```diff
@@ -428,6 +450,12 @@ std::unique_ptr<ZoneInfoSource> FileZoneInfoSource::Open(
   // Use of the "file:" prefix is intended for testing purposes only.
   const std::size_t pos = (name.compare(0, 5, "file:") == 0) ? 5 : 0;
 
+  // Reject path-traversal components so that a relative zone name cannot
+  // escape the zoneinfo directory (e.g., "../../etc/passwd").
+  if (name.find("..", pos) != std::string::npos) {
```
devbww: Finding any `".."` is too restrictive (e.g., `"x..y"`).

src/time_zone_lookup_test.cc
```diff
@@ -173,6 +173,21 @@ TEST(TimeZone, Failures) {
   EXPECT_FALSE(load_time_zone("", &tz));
   EXPECT_EQ(chrono::system_clock::from_time_t(0),
             convert(civil_second(1970, 1, 1, 0, 0, 0), tz));  // UTC
+
+  // Reject path-traversal components.
+  tz = LoadZone("America/Los_Angeles");
```
devbww: You don't need to set `tz` to a non-UTC zone if you're not going to test that it is UTC after the failing `load_time_zone()` (et seq).

src/time_zone_lookup_test.cc
```diff
@@ -173,6 +173,21 @@ TEST(TimeZone, Failures) {
...
+  EXPECT_FALSE(load_time_zone("file:../etc/passwd", &tz));
+  EXPECT_FALSE(load_time_zone("file:../../etc/passwd", &tz));
+  EXPECT_FALSE(load_time_zone("file:/../etc/passwd", &tz));
+
+  // Reject non-regular files.
+  tz = LoadZone("America/Los_Angeles");
+  EXPECT_FALSE(load_time_zone("file:/dev/null", &tz));
+  EXPECT_FALSE(load_time_zone("file:/dev/stdin", &tz));
```
devbww, line 185: Do we care that these will probably fail in non-unix environments for reasons other than path and file-type restrictions?

src/time_zone_info.cc
```diff
@@ -386,8 +392,24 @@ inline FilePtr FOpen(const char* path, const char* mode) {
   if (fopen_s(&fp, path, mode) != 0) fp = nullptr;
   return FilePtr(fp, fclose);
 #else
-  // TODO: Enable the close-on-exec flag.
-  return FilePtr(fopen(path, mode), fclose);
+  // Open non-blocking and verify the target is a regular file before handing it
+  // to stdio. Zone names are potentially attacker-controlled, and a plain
+  // fopen() on a FIFO or device node (reachable via the "file:" prefix or an
+  // absolute path) would block indefinitely or read unbounded data.
+  const int fd = open(path, O_RDONLY | O_NONBLOCK | O_CLOEXEC);
```
devbww, line 399: While we actually only call `FOpen()` with a mode of `"rb"`, the `O_RDONLY` assumes it.

src/time_zone_info.cc
```diff
@@ -386,8 +392,24 @@ inline FilePtr FOpen(const char* path, const char* mode) {
...
+  struct stat st;
+  if (fstat(fd, &st) != 0 || !S_ISREG(st.st_mode)) {
+    close(fd);
+    return FilePtr(nullptr, fclose);
+  }
+  FILE* fp = fdopen(fd, mode);
+  if (fp == nullptr) {
+    close(fd);
+  }
+  return FilePtr(fp, fclose);
```
devbww: I would tend to write all this with single success and failure/cleanup paths.  That is, ...
```
  if (fd >= 0) {
    stat st;
    if (fstat(fd, &st) == 0 && S_ISREG(st.st_mode)) {
      FILE* fp = fdopen(fd, mode);
      if (fp != nullptr) return FilePtr(fp, fclose);
    }
    close(fd);
  }
  return FilePtr(nullptr, fclose);
```

src/time_zone_info.cc
derekmauro: Done.

src/time_zone_lookup_test.cc
derekmauro: Done.

src/time_zone_info.cc
derekmauro: Done.

src/time_zone_lookup_test.cc
derekmauro: This seems good enough for now.

src/time_zone_info.cc
```diff
@@ -379,18 +385,46 @@ namespace {
...
+// Returns true if the zone name contains a ".." path-traversal component.
+inline bool HasPathTraversal(const std::string& name, std::size_t pos) {
+  // Exact match: ".."
+  if (name.compare(pos, std::string::npos, "..") == 0) return true;
+  // Leading component: "../"
+  if (name.compare(pos, 3, "../") == 0) return true;
+  // Interior component: "/../"
+  if (name.find("/../", pos) != std::string::npos) return true;
+  // Trailing component: "/.."
+  if (name.size() >= pos + 3 &&
```
devbww: Because the only thing we know is `pos <= name.size()`, I'd write this as
```
  if (name.size() - pos >= 3 && ...
```
to avoid even the odor of overflow.

src/time_zone_info.cc
```diff
@@ -379,18 +385,46 @@ namespace {
...
+      FILE* fp = fdopen(fd, "rb");
+      if (fp != nullptr) return FilePtr(fp, fclose);
+    }
+    close(fd);
+  }
+  return FilePtr(nullptr, fclose);
 #endif
 }
 
+// Returns true if the zone name contains a ".." path-traversal component.
```
devbww: Mention `pos`?

src/time_zone_info.cc
```diff
@@ -379,18 +385,46 @@ namespace {
...
+inline bool HasPathTraversal(const std::string& name, std::size_t pos) {
```
devbww: I'm not sure I like the term `HasPathTraversal()`.  Any path can be traversed.

How about something like `UnsafePath()`, which could then be expanded to cases other than `"..`"?

src/time_zone_info.cc
derekmauro: Done.

src/time_zone_info.cc
derekmauro: Done.

src/time_zone_info.cc
derekmauro: Done.

devbww (approved):

mkruskal-google (approved):

PR #354, opened 2026-07-17
Fix overflow/underflow detection in join_seconds

### LINKED ISSUE #199: Detect if include/cctz/time_zone.h:join_seconds() would overflow the output time_point
```
template <typename Rep, std::intmax_t Denom>
bool join_seconds(
    const time_point<seconds>& sec, const femtoseconds& fs,
    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp)
```
needs to return `false` when its result is unrepresentable, and without invoking UB.

derekmauro:
Improve the precision of overflow and underflow checks in the `join_seconds` template function:
  - Use `std::intmax_t` for intermediate subsecond calculations to prevent premature overflow when the target representation type is small.
  - Refine the underflow check to correctly handle boundary cases near the minimum limit, avoiding unsafe subtraction.
  - Handle cases where the duration denominator is larger than the maximum limit of the representation type.
  - Support floating-point duration representations.
  - Reduce duration periods to lowest terms via `typename Period::type` in `cctz::parse()` to support non-reduced ratios.
  - Align `detail::split_seconds()` with floor semantics for pre-epoch times.

Fixes #199

### DIFF
--- a/include/cctz/time_zone.h
+++ b/include/cctz/time_zone.h
@@ -404,9 +404,36 @@ bool join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
     time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
   using D = std::chrono::duration<Rep, std::ratio<1, Denom>>;
-  // TODO(#199): Return false if result unrepresentable as a time_point<D>.
-  *tpp = std::chrono::time_point_cast<D>(sec);
-  *tpp += std::chrono::duration_cast<D>(fs);
+  // Use a 64-bit representation for intermediate subsecond calculations
+  // to avoid premature overflow if Rep is a smaller type (e.g., int8_t).
+  using D_check = std::chrono::duration<std::int64_t, std::ratio<1, Denom>>;
+  auto count = sec.time_since_epoch().count();
+  auto sub = std::chrono::duration_cast<D_check>(fs).count();  // [0, Denom)
+
+  // Check for overflow.
+  if (sub > (std::numeric_limits<Rep>::max)()) {
+    // If subseconds alone exceed the max representable value, any non-negative
+    // seconds count will cause overflow. Negative seconds might bring it back
+    // into range, which is handled by the underflow check below.
+    if (count >= 0) return false;
+  } else {
+    // Equivalent to: count * Denom + sub > max, but safe from overflow.
+    if (count > ((std::numeric_limits<Rep>::max)() - sub) / Denom) return false;
+  }
+
+  // Check for underflow.
+  // Equivalent to: count * Denom + sub < min, but safe from underflow.
+  // We cannot use "min - sub" directly as it would underflow.
+  const auto min_div = (std::numeric_limits<Rep>::min)() / Denom;
+  if (count < min_div) {
+    // If count is less than min_div - 1, it's definitely underflow.
+    if (count < min_div - 1) return false;
+    // If count is exactly min_div - 1, we must check if subseconds are
+    // sufficient to bring the total value back above min.
+    const auto min_mod = (std::numeric_limits<Rep>::min)() % Denom;
+    if (sub < Denom + min_mod) return false;
+  }
+  *tpp = time_point<D>() + D{static_cast<Rep>(count * Denom + sub)};
   return true;
 }
 
--- a/include/cctz/time_zone.h
+++ b/include/cctz/time_zone.h
@@ -25,6 +25,7 @@
 #include <limits>
 #include <ratio>
 #include <string>
+#include <type_traits>
 #include <utility>
 
 #include "cctz/civil_time.h"
@@ -402,13 +403,13 @@ inline std::pair<time_point<seconds>, seconds> split_seconds(
 template <typename Rep, std::intmax_t Denom>
 bool join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
-    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
+    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp,
+    std::true_type /* is_integral */) {
   using D = std::chrono::duration<Rep, std::ratio<1, Denom>>;
-  // Use a 64-bit representation for intermediate subsecond calculations
-  // to avoid premature overflow if Rep is a smaller type (e.g., int8_t).
-  using D_check = std::chrono::duration<std::int64_t, std::ratio<1, Denom>>;
-  auto count = sec.time_since_epoch().count();
-  auto sub = std::chrono::duration_cast<D_check>(fs).count();  // [0, Denom)
+  using D_check = std::chrono::duration<std::intmax_t, std::ratio<1, Denom>>;
+  const auto count = static_cast<std::intmax_t>(sec.time_since_epoch().count());
+  const auto sub =
+      std::chrono::duration_cast<D_check>(fs).count();  // [0, Denom)
 
   // Check for overflow.
   if (sub > (std::numeric_limits<Rep>::max)()) {
@@ -422,19 +423,40 @@ bool join_seconds(
   }
 
   // Check for underflow.
-  // Equivalent to: count * Denom + sub < min, but safe from underflow.
-  // We cannot use "min - sub" directly as it would underflow.
-  const auto min_div = (std::numeric_limits<Rep>::min)() / Denom;
+  // Equivalent to: count * Denom + sub < lowest, but safe from underflow.
+  // We cannot use "lowest - sub" directly as it would underflow.
+  const auto min_div = (std::numeric_limits<Rep>::lowest)() / Denom;
   if (count < min_div) {
     // If count is less than min_div - 1, it's definitely underflow.
     if (count < min_div - 1) return false;
     // If count is exactly min_div - 1, we must check if subseconds are
-    // sufficient to bring the total value back above min.
-    const auto min_mod = (std::numeric_limits<Rep>::min)() % Denom;
+    // sufficient to bring the total value back above lowest.
+    const auto min_mod = (std::numeric_limits<Rep>::lowest)() % Denom;
     if (sub < Denom + min_mod) return false;
+    *tpp =
+        time_point<D>() + D{static_cast<Rep>(min_div * Denom + (sub - Denom))};
+    return true;
   }
   *tpp = time_point<D>() + D{static_cast<Rep>(count * Denom + sub)};
   return true;
+}
+
+template <typename Rep, std::intmax_t Denom>
+bool join_seconds(
+    const time_point<seconds>& sec, const femtoseconds& fs,
+    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp,
+    std::false_type /* is_integral */) {
+  using D = std::chrono::duration<Rep, std::ratio<1, Denom>>;
+  *tpp = std::chrono::time_point_cast<D>(sec);
+  *tpp += std::chrono::duration_cast<D>(fs);
+  return true;
+}
+
+template <typename Rep, std::intmax_t Denom>
+bool join_seconds(
+    const time_point<seconds>& sec, const femtoseconds& fs,
+    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
+  return join_seconds(sec, fs, tpp, std::is_integral<Rep>());
 }
 
 template <typename Rep, std::intmax_t Num>
@@ -450,7 +472,8 @@ bool join_seconds(
     count -= 1;
   }
   if (count > (std::numeric_limits<Rep>::max)()) return false;
-  if (count < (std::numeric_limits<Rep>::min)()) return false;
+  if (count < (std::numeric_limits<Rep>::lowest)())
+    return false;
   *tpp = time_point<D>() + D{static_cast<Rep>(count)};
   return true;
 }
@@ -462,7 +485,8 @@ bool join_seconds(
   using D = std::chrono::duration<Rep, std::ratio<1, 1>>;
   auto count = sec.time_since_epoch().count();
   if (count > (std::numeric_limits<Rep>::max)()) return false;
-  if (count < (std::numeric_limits<Rep>::min)()) return false;
+  if (count < (std::numeric_limits<Rep>::lowest)())
+    return false;
   *tpp = time_point<D>() + D{static_cast<Rep>(count)};
   return true;
 }
--- a/include/cctz/time_zone.h
+++ b/include/cctz/time_zone.h
@@ -396,12 +396,8 @@ namespace detail {
 template <typename D>
 std::pair<time_point<seconds>, D> split_seconds(const time_point<D>& tp) {
   auto sec = std::chrono::time_point_cast<seconds>(tp);
-  auto sub = tp - sec;
-  if (sub < D::zero()) {
-    sec -= seconds(1);
-    sub += seconds(1);
-  }
-  return {sec, std::chrono::duration_cast<D>(sub)};
+  if (sec > tp) sec -= seconds{1};  // TODO(C++17): use std::chrono::floor
+  return {sec, std::chrono::duration_cast<D>(tp - sec)};
 }
 
 inline std::pair<time_point<seconds>, seconds> split_seconds(
@@ -492,7 +488,7 @@ join_seconds(
   auto count = sec.time_since_epoch().count();
   if (count > (std::numeric_limits<Rep>::max)()) return false;
   if (count < (std::numeric_limits<Rep>::lowest)()) return false;
-  *tpp = std::chrono::time_point_cast<D>(sec);
+  *tpp = time_point<D>() + D{static_cast<Rep>(count)};
   return true;
 }
 
--- a/include/cctz/time_zone.h
+++ b/include/cctz/time_zone.h
@@ -472,8 +472,7 @@ bool join_seconds(
     count -= 1;
   }
   if (count > (std::numeric_limits<Rep>::max)()) return false;
-  if (count < (std::numeric_limits<Rep>::lowest)())
-    return false;
+  if (count < (std::numeric_limits<Rep>::lowest)()) return false;
   *tpp = time_point<D>() + D{static_cast<Rep>(count)};
   return true;
 }
@@ -485,8 +484,7 @@ bool join_seconds(
   using D = std::chrono::duration<Rep, std::ratio<1, 1>>;
   auto count = sec.time_since_epoch().count();
   if (count > (std::numeric_limits<Rep>::max)()) return false;
-  if (count < (std::numeric_limits<Rep>::lowest)())
-    return false;
+  if (count < (std::numeric_limits<Rep>::lowest)()) return false;
   *tpp = time_point<D>() + D{static_cast<Rep>(count)};
   return true;
 }
--- a/include/cctz/time_zone.h
+++ b/include/cctz/time_zone.h
@@ -267,15 +267,20 @@ std::string format(const std::string&, const time_point<seconds>&,
 bool parse(const std::string&, const std::string&, const time_zone&,
            time_point<seconds>*, femtoseconds*, std::string* err = nullptr);
 template <typename Rep, std::intmax_t Denom>
-bool join_seconds(
+typename std::enable_if<std::is_integral<Rep>::value, bool>::type join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
     time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp);
+template <typename Rep, std::intmax_t Num, std::intmax_t Denom>
+typename std::enable_if<std::is_floating_point<Rep>::value, bool>::type
+join_seconds(
+    const time_point<seconds>& sec, const femtoseconds& fs,
+    time_point<std::chrono::duration<Rep, std::ratio<Num, Denom>>>* tpp);
 template <typename Rep, std::intmax_t Num>
-bool join_seconds(
+typename std::enable_if<std::is_integral<Rep>::value, bool>::type join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
     time_point<std::chrono::duration<Rep, std::ratio<Num, 1>>>* tpp);
 template <typename Rep>
-bool join_seconds(
+typename std::enable_if<std::is_integral<Rep>::value, bool>::type join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
     time_point<std::chrono::duration<Rep, std::ratio<1, 1>>>* tpp);
 bool join_seconds(const time_point<seconds>& sec, const femtoseconds&,
@@ -365,13 +370,17 @@ inline std::string format(const std::string& fmt, const time_point<D>& tp,
 //   if (cctz::parse("%Y-%m-%d", "2015-10-09", tz, &tp)) {
 //     ...
 //   }
-template <typename D>
+template <typename Rep, typename Period>
 inline bool parse(const std::string& fmt, const std::string& input,
-                  const time_zone& tz, time_point<D>* tpp) {
+                  const time_zone& tz,
+                  time_point<std::chrono::duration<Rep, Period>>* tpp) {
   time_point<seconds> sec;
   detail::femtoseconds fs;
-  return detail::parse(fmt, input, tz, &sec, &fs) &&
-         detail::join_seconds(sec, fs, tpp);
+  if (!detail::parse(fmt, input, tz, &sec, &fs)) return false;
+  time_point<std::chrono::duration<Rep, typename Period::type>> tp;
+  if (!detail::join_seconds(sec, fs, &tp)) return false;
+  *tpp = tp;
+  return true;
 }
 
 namespace detail {
@@ -401,13 +410,12 @@ inline std::pair<time_point<seconds>, seconds> split_seconds(
 // Floors to the resolution of time_point<D>. Returns false if time_point<D>
 // is not of sufficient range.
 template <typename Rep, std::intmax_t Denom>
-bool join_seconds(
+typename std::enable_if<std::is_integral<Rep>::value, bool>::type join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
-    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp,
-    std::true_type /* is_integral */) {
+    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
   using D = std::chrono::duration<Rep, std::ratio<1, Denom>>;
   using D_check = std::chrono::duration<std::intmax_t, std::ratio<1, Denom>>;
-  const auto count = static_cast<std::intmax_t>(sec.time_since_epoch().count());
+  const std::intmax_t count = sec.time_since_epoch().count();
   const auto sub =
       std::chrono::duration_cast<D_check>(fs).count();  // [0, Denom)
 
@@ -441,26 +449,19 @@ bool join_seconds(
   return true;
 }
 
-template <typename Rep, std::intmax_t Denom>
-bool join_seconds(
+template <typename Rep, std::intmax_t Num, std::intmax_t Denom>
+typename std::enable_if<std::is_floating_point<Rep>::value, bool>::type
+join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
-    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp,
-    std::false_type /* is_integral */) {
-  using D = std::chrono::duration<Rep, std::ratio<1, Denom>>;
+    time_point<std::chrono::duration<Rep, std::ratio<Num, Denom>>>* tpp) {
+  using D = std::chrono::duration<Rep, std::ratio<Num, Denom>>;
   *tpp = std::chrono::time_point_cast<D>(sec);
   *tpp += std::chrono::duration_cast<D>(fs);
   return true;
 }
 
-template <typename Rep, std::intmax_t Denom>
-bool join_seconds(
-    const time_point<seconds>& sec, const femtoseconds& fs,
-    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
-  return join_seconds(sec, fs, tpp, std::is_integral<Rep>());
-}
-
 template <typename Rep, std::intmax_t Num>
-bool join_seconds(
+typename std::enable_if<std::is_integral<Rep>::value, bool>::type join_seconds(
     const time_point<seconds>& sec, const femtoseconds&,
     time_point<std::chrono::duration<Rep, std::ratio<Num, 1>>>* tpp) {
   using D = std::chrono::duration<Rep, std::ratio<Num, 1>>;
@@ -478,14 +479,14 @@ bool join_seconds(
 }
 
 template <typename Rep>
-bool join_seconds(
+typename std::enable_if<std::is_integral<Rep>::value, bool>::type join_seconds(
     const time_point<seconds>& sec, const femtoseconds&,
     time_point<std::chrono::duration<Rep, std::ratio<1, 1>>>* tpp) {
   using D = std::chrono::duration<Rep, std::ratio<1, 1>>;
   auto count = sec.time_since_epoch().count();
   if (count > (std::numeric_limits<Rep>::max)()) return false;
   if (count < (std::numeric_limits<Rep>::lowest)()) return false;
-  *tpp = time_point<D>() + D{static_cast<Rep>(count)};
+  *tpp = std::chrono::time_point_cast<D>(sec);
   return true;
 }
 
--- a/include/cctz/time_zone.h
+++ b/include/cctz/time_zone.h
@@ -267,7 +267,8 @@ std::string format(const std::string&, const time_point<seconds>&,
 bool parse(const std::string&, const std::string&, const time_zone&,
            time_point<seconds>*, femtoseconds*, std::string* err = nullptr);
 template <typename Rep, std::intmax_t Denom>
-typename std::enable_if<std::is_integral<Rep>::value, bool>::type join_seconds(
+typename std::enable_if<std::is_integral<Rep>::value, bool>::type
+join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
     time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp);
 template <typename Rep, std::intmax_t Num, std::intmax_t Denom>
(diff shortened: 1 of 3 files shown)


devbww:
@derekmauro ... Just a heads-up that I think I'll have some comments for this, but it might take me a day or two to formulate them.  Thanks.

derekmauro:
@devbww - No problem. If you haven't figured it out yet, thanks to Mythos we are all inundated with bugs, and apparently I am not the only person with access. Take your time.

devbww:
> If you haven't figured it out yet, thanks to Mythos we are all inundated with bugs, and apparently I am not the only person with access.

I had not figured it out, but I was beginning to wonder.  Hopefully the pickings are getting slimmer.  :-)

derekmauro:
@devbww - Do you still want to review this one?

devbww:
> @devbww - Do you still want to review this one?

Yes.  I'm sorry for the delay.  That bout of recent PRs sucked up my free time.  I'll respond today.  Thanks.

include/cctz/time_zone.h
```diff
@@ -404,9 +404,36 @@ bool join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
     time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
   using D = std::chrono::duration<Rep, std::ratio<1, Denom>>;
-  // TODO(#199): Return false if result unrepresentable as a time_point<D>.
-  *tpp = std::chrono::time_point_cast<D>(sec);
-  *tpp += std::chrono::duration_cast<D>(fs);
+  // Use a 64-bit representation for intermediate subsecond calculations
+  // to avoid premature overflow if Rep is a smaller type (e.g., int8_t).
+  using D_check = std::chrono::duration<std::int64_t, std::ratio<1, Denom>>;
```
devbww: We should avoid `std::int64_t`.  It isn't even guaranteed to exist.  We don't use the fixed-size integer types anywhere in cctz except in these overflow tests (which seems necessary to make tight expectations, but which would be nice to avoid), the fuzz tests (which I disavow, but should probably be changed), and the Windows time-zone-name stuff (which I assume is required by the API).

So, `std::int_fast64_t` would be a better choice.  That will also make `count` and `sub` have the same type, which looks like a feature.

That said, `std::intmax_t` is probably the best choice, and that is the type of `Denom` anyway (and it might be also be `Rep`).

And that said, `Rep` may also be a floating-point type, so we'll need to be careful there too (see below).

include/cctz/time_zone.h
```diff
@@ -404,9 +404,36 @@ bool join_seconds(
...
+    if (count >= 0) return false;
+  } else {
+    // Equivalent to: count * Denom + sub > max, but safe from overflow.
+    if (count > ((std::numeric_limits<Rep>::max)() - sub) / Denom) return false;
+  }
+
+  // Check for underflow.
+  // Equivalent to: count * Denom + sub < min, but safe from underflow.
+  // We cannot use "min - sub" directly as it would underflow.
+  const auto min_div = (std::numeric_limits<Rep>::min)() / Denom;
```
devbww: `std::numeric_limits<Rep>::min()` is not the "minimum" value of `Rep` when it is floating point.  Rather, it is the smallest, positive, normalized value.  So, we'll either have to avoid `min()` or treat floating point separately.

Note that we similarly use `std::numeric_limits<Rep>::min()` in a couple of the other existing `join_seconds()` overloads, so we should deal with them somehow too.

I don't think we can now simply exclude floating-point representations, despite them currently misbehaving on pre-epoch times.

include/cctz/time_zone.h
```diff
@@ -404,9 +404,36 @@ bool join_seconds(
...
+  if (count < min_div) {
+    // If count is less than min_div - 1, it's definitely underflow.
+    if (count < min_div - 1) return false;
+    // If count is exactly min_div - 1, we must check if subseconds are
+    // sufficient to bring the total value back above min.
+    const auto min_mod = (std::numeric_limits<Rep>::min)() % Denom;
+    if (sub < Denom + min_mod) return false;
+  }
+  *tpp = time_point<D>() + D{static_cast<Rep>(count * Denom + sub)};
```
devbww, line 450: Until C++20, there is no guarantee that the `std::chrono::system_clock` epoch is the same as the Unix epoch, so carefully checking that `count * Denom + sub` fits in `Rep` may all be for naught when chrono includes some additional offset in its calculations.

But that said, I'd be happy to place something along the lines of ...
```
#if __cplusplus < 202002L
  // Assert that the std::chrono::system_clock epoch is the Unix epoch.
  // This is required by C++20, but was also ubiquitous before that.
  // We use this in join_seconds() to simplify overflow detection.
  assert(std::chrono::system_clock::from_time_t(0).time_since_epoch() ==
         std::chrono::system_clock::duration::zero());
#endif
```
in, say, `detail::parse()` so that we might assume there is no such offset.

src/time_zone_format_test.cc
```diff
@@ -1628,52 +1628,84 @@ TEST(Parse, MaxRange) {
 TEST(Parse, TimePointOverflow) {
   const time_zone utc = utc_time_zone();
 
+  // Test with nanosecond resolution (typical 64-bit time_point).
   using D = chrono::duration<std::int64_t, std::nano>;
```
devbww: Is there any reason to test `std::femto` or beyond?

src/time_zone_format_test.cc
```diff
@@ -1628,52 +1628,84 @@ TEST(Parse, MaxRange) {
...
 
+  // Max representable time_point<D> is 2262-04-11T23:47:16.854775807 UTC.
   EXPECT_TRUE(
       parse(RFC3339_full, "2262-04-11T23:47:16.8547758079+00:00", utc, &tp));
   EXPECT_EQ(tp, time_point<D>::max());
   EXPECT_EQ("2262-04-11T23:47:16.854775807+00:00",
             cctz::format(RFC3339_full, tp, utc));
-#if 0
-  // TODO(#199): Will fail until cctz::parse() properly detects overflow.
+
```
devbww: Trailing whitespace.

include/cctz/time_zone.h
derekmauro: Done

include/cctz/time_zone.h
derekmauro: Fixed to use `std::numeric_limits<Rep>::lowest()`. Added tag dispatch for floating point types.

include/cctz/time_zone.h
derekmauro: Done

src/time_zone_format_test.cc
derekmauro: Added a bunch of tests.

src/time_zone_format_test.cc
derekmauro: Fixed

include/cctz/time_zone.h
```diff
@@ -402,14 +403,62 @@ inline std::pair<time_point<seconds>, seconds> split_seconds(
 template <typename Rep, std::intmax_t Denom>
 bool join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
-    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
+    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp,
+    std::true_type /* is_integral */) {
```
devbww: Consider changing the comment here and below to `/* is_integral<Rep> */` to make it clear what type we're talking about.

**But**, I think the next suggestion is better, and avoids this altogether.

include/cctz/time_zone.h
```diff
@@ -402,14 +403,62 @@ inline std::pair<time_point<seconds>, seconds> split_seconds(
...
+  using D = std::chrono::duration<Rep, std::ratio<1, Denom>>;
+  using D_check = std::chrono::duration<std::intmax_t, std::ratio<1, Denom>>;
+  const auto count = static_cast<std::intmax_t>(sec.time_since_epoch().count());
```
devbww: It seems clearer and safer to do simple initialization with a named type rather than static_cast'ing into an auto.  That is, ...
```
  const std::intmax_t count = sec.time_since_epoch().count();
```

include/cctz/time_zone.h
```diff
@@ -402,14 +403,62 @@ inline std::pair<time_point<seconds>, seconds> split_seconds(
...
   *tpp += std::chrono::duration_cast<D>(fs);
   return true;
 }
 
+template <typename Rep, std::intmax_t Denom>
+bool join_seconds(
+    const time_point<seconds>& sec, const femtoseconds& fs,
+    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
+  return join_seconds(sec, fs, tpp, std::is_integral<Rep>());
+}
```
devbww: I would avoid this template that indirects to the other, now true/false_type-overloaded templates by simply making the original "1/Denom" template, and the new floating-point one, only accept the desired type.

That is, remove this helper and introduce `std::enable_if<>` into the others as ...
```
template <typename Rep, std::intmax_t Denom>
typename std::enable_if<std::is_integral<Rep>::value, bool>::type
join_seconds(
    const time_point<seconds>& sec, const femtoseconds& fs,
    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp);
```
and
```
template <typename Rep, std::intmax_t Denom>
typename std::enable_if<std::is_floating_point<Rep>::value, bool>::type
join_seconds(
    const time_point<seconds>& sec, const femtoseconds& fs,
    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp);
```
This also makes it very clear that the second one is for `is_floating_point<Rep>`, and not just `!is_integral<Rep>`, even if we might know from elsewhere that `Rep` has to be arithmetic.

**But**, also see below about generalizing the floating-point version.

include/cctz/time_zone.h
```diff
@@ -402,14 +403,62 @@ inline std::pair<time_point<seconds>, seconds> split_seconds(
...
+    return true;
+  }
+  *tpp = time_point<D>() + D{static_cast<Rep>(count * Denom + sub)};
+  return true;
+}
+
+template <typename Rep, std::intmax_t Denom>
+bool join_seconds(
+    const time_point<seconds>& sec, const femtoseconds& fs,
+    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp,
```
devbww: We don't want to limit this new floating-point version to only representations of sub-seconds, however.  So, ...
```
template <typename Rep, std::intmax_t Num, std::intmax_t Denom>
typename std::enable_if<std::is_floating_point<Rep>::value, bool>::type
join_seconds(
    const time_point<seconds>& sec, const femtoseconds& fs,
    time_point<std::chrono::duration<Rep, std::ratio<Num, Denom>>>* tpp) {
  using D = std::chrono::duration<Rep, std::ratio<Num, Denom>>;
  *tpp = std::chrono::time_point_cast<D>(sec);
  *tpp += std::chrono::duration_cast<D>(fs);
  return true;
}
```

Methinks all the other `join_seconds()` templates should then be restricted to `is_integral<Rep>`.

include/cctz/time_zone.h
```diff
@@ -402,14 +403,62 @@ inline std::pair<time_point<seconds>, seconds> split_seconds(
 template <typename Rep, std::intmax_t Denom>
 bool join_seconds(
     const time_point<seconds>& sec, const femtoseconds& fs,
-    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp) {
+    time_point<std::chrono::duration<Rep, std::ratio<1, Denom>>>* tpp,
```
devbww: Aside:  We should probably do something using `std::ratio<>::type` to ensure we're dealing with the lowest terms by the time we reach for these templates.

include/cctz/time_zone.h
```diff
@@ -402,14 +403,62 @@ inline std::pair<time_point<seconds>, seconds> split_seconds(
...
+    std::true_type /* is_integral */) {
+  using D = std::chrono::duration<Rep, std::ratio<1, Denom>>;
+  using D_check = std::chrono::duration<std::intmax_t, std::ratio<1, Denom>>;
+  const auto count = static_cast<std::intmax_t>(sec.time_since_epoch().count());
+  const auto sub =
+      std::chrono::duration_cast<D_check>(fs).count();  // [0, Denom)
```
devbww, line 422: Aside:  Hopefully chrono is smart enough to avoid unnecessary overflow when converting to a ratio with a very large Denom.

src/time_zone_format_test.cc
```diff
@@ -1628,52 +1628,144 @@ TEST(Parse, MaxRange) {
...
+  EXPECT_FALSE(parse(RFC3339_full, "1970-01-01T00:00:10+00:00", utc, &atp));
+
+  // -9 seconds fits in 64-bit attoseconds.
+  EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:59:51+00:00", utc, &atp));
+  EXPECT_EQ(atp.time_since_epoch().count(), -9000000000000000000LL);
+  // -10 seconds underflows 64-bit attoseconds limit.
+  EXPECT_FALSE(parse(RFC3339_full, "1969-12-31T23:59:50+00:00", utc, &atp));
+
+  // Test with 1-second resolution using int8_t (very narrow range: [-128, 127] seconds).
   using DS = chrono::duration<std::int8_t, chrono::seconds::period>;
```
devbww: Elsewhere we omit the Period when it is the default `std::ratio<1, 1>` (e.g., in the definition of `cctz::seconds`).  My suggestion would be to keep doing that here.

src/time_zone_format_test.cc
```diff
@@ -1628,52 +1628,144 @@ TEST(Parse, MaxRange) {
...
 
+  // Min: -128 seconds.
   EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:57:52+00:00", utc, &stp));
   EXPECT_EQ(stp, time_point<DS>::min());
   EXPECT_EQ("1969-12-31T23:57:52+00:00", cctz::format(RFC3339_full, stp, utc));
+  // -129 seconds should fail.
   EXPECT_FALSE(parse(RFC3339_full, "1969-12-31T23:57:51.9+00:00", utc, &stp));
 
+  // Test with 1-minute resolution using int8_t (range: [-128, 127] minutes).
   using DM = chrono::duration<std::int8_t, chrono::minutes::period>;
```
devbww: We must use a named ratio like `std::femto` when there is no matching `chrono::duration` to extract a period from, so I would suggest always using ratios directly and saying `std::ratio<60>` here (which requires `#include <ratio>`).  But, it's a small point.

src/time_zone_format_test.cc
```diff
@@ -1628,52 +1628,144 @@ TEST(Parse, MaxRange) {
 TEST(Parse, TimePointOverflow) {
```
devbww, line 1637: All these cases really have the same form, so I'd make them look as similar as possible.

I'd also nest each case so we don't have to conjure up new type and variable names each time.  (Perhaps they could even be separate tests.)

Anyway, to avoid you having to guess everything, here is what I mean (despite its length):
```
TEST(Parse, TimePointOverflow) {
  const time_zone utc = utc_time_zone();
  {
    // Test with nanosecond resolution (typical 64-bit time_point).
    using D = chrono::duration<std::int64_t, std::nano>;
    time_point<D> tp;

    // Max time_point<D> is 2262-04-11T23:47:16.854775807+00:00.
    EXPECT_TRUE(
        parse(RFC3339_full, "2262-04-11T23:47:16.8547758079+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::max());
    EXPECT_EQ("2262-04-11T23:47:16.854775807+00:00",
              cctz::format(RFC3339_full, tp, utc));
    // 1 nanosecond beyond max should fail.
    EXPECT_FALSE(
        parse(RFC3339_full, "2262-04-11T23:47:16.8547758080+00:00", utc, &tp));

    // Min time_point<D> is 1677-09-21T00:12:43.145224192+00:00.
    EXPECT_TRUE(
        parse(RFC3339_full, "1677-09-21T00:12:43.1452241920+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::min());
    EXPECT_EQ("1677-09-21T00:12:43.145224192+00:00",
              cctz::format(RFC3339_full, tp, utc));
    // 1 nanosecond below min should fail.
    EXPECT_FALSE(
        parse(RFC3339_full, "1677-09-21T00:12:43.1452241919+00:00", utc, &tp));
  }
  {
    // Test with femtosecond resolution.
    using D = chrono::duration<std::int64_t, std::femto>;
    time_point<D> tp;

    // Max time_point<D> is 1970-01-01T02:33:43.372036854775807+00:00.
    EXPECT_TRUE(parse(RFC3339_full,
                      "1970-01-01T02:33:43.3720368547758079+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::max());
    EXPECT_EQ("1970-01-01T02:33:43.372036854775807+00:00",
              cctz::format(RFC3339_full, tp, utc));
    // 1 femtosecond beyond max should fail.
    EXPECT_FALSE(parse(RFC3339_full,
                       "1970-01-01T02:33:43.3720368547758080+00:00", utc, &tp));

    // Min time_point<D> is 1969-12-31T21:26:16.627963145224192+00:00.
    EXPECT_TRUE(parse(RFC3339_full,
                      "1969-12-31T21:26:16.6279631452241920+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::min());
    EXPECT_EQ("1969-12-31T21:26:16.627963145224192+00:00",
              cctz::format(RFC3339_full, tp, utc));
    // 1 femtosecond below min should fail.
    EXPECT_FALSE(parse(RFC3339_full,
                       "1969-12-31T21:26:16.6279631452241919+00:00", utc, &tp));
  }
  {
    // Test with attosecond resolution.
    using D = chrono::duration<std::int64_t, std::atto>;
    time_point<D> tp;

    // Max time_point<D> is 1970-01-01T00:00:09.223372036854775807+00:00,
    // but cctz::parse() truncates (towards zero) to a femtosecond boundary,
    // so the last three decimal places are lost.
    EXPECT_TRUE(parse(RFC3339_full,
                      "1970-01-01T00:00:09.223372036854775807+00:00", utc,
                      &tp));
    EXPECT_EQ(tp.time_since_epoch(), D{9223372036854775000LL});
    EXPECT_EQ("1970-01-01T00:00:09.223372036854775+00:00",
              cctz::format(RFC3339_full, tp, utc));
    // Moving into the next femtosecond should fail.
    EXPECT_FALSE(parse(RFC3339_full,
                       "1970-01-01T00:00:09.223372036854776000+00:00", utc,
                       &tp));

    // Min time_point<D> is 1969-12-31T23:59:50.776627963145224192+00:00,
    // but cctz::parse() truncates (towards zero) to a femtosecond boundary,
    // so we must round up.
    EXPECT_TRUE(parse(RFC3339_full,
                      "1969-12-31T23:59:50.776627963145225000+00:00", utc,
                      &tp));
    EXPECT_EQ(tp.time_since_epoch(), D{-9223372036854775000LL});
    EXPECT_EQ("1969-12-31T23:59:50.776627963145225+00:00",
              cctz::format(RFC3339_full, tp, utc));
    // 1 attosecond below min should fail.
    EXPECT_FALSE(parse(RFC3339_full,
                       "1969-12-31T23:59:50.776627963145224999+00:00", utc,
                       &tp));
  }
  {
    // Test with 1-second resolution using int8_t ([-128, 127] seconds).
    using D = chrono::duration<std::int8_t>;
    time_point<D> tp;

    // Max time_point<D> is 1970-01-01T00:02:07+00:00.
    EXPECT_TRUE(parse(RFC3339_full, "1970-01-01T00:02:07.9+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::max());
    EXPECT_EQ("1970-01-01T00:02:07+00:00", cctz::format(RFC3339_full, tp, utc));
    // 1 second beyond max should fail.
    EXPECT_FALSE(parse(RFC3339_full, "1970-01-01T00:02:08+00:00", utc, &tp));

    // Min time_point<D> is 1969-12-31T23:57:52+00:00.
    EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:57:52+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::min());
    EXPECT_EQ("1969-12-31T23:57:52+00:00", cctz::format(RFC3339_full, tp, utc));
    // 1 second below min should fail.
    EXPECT_FALSE(parse(RFC3339_full, "1969-12-31T23:57:51+00:00", utc, &tp));
  }
  {
    // Test with 1-minute resolution using int8_t ([-128, 127] minutes).
    using D = chrono::duration<std::int8_t, std::ratio<60>>;
    time_point<D> tp;

    // Max time_point<D> is 1970-01-01T02:07:00+00:00.
    EXPECT_TRUE(parse(RFC3339_full, "1970-01-01T02:07:00+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::max());
    EXPECT_EQ("1970-01-01T02:07:00+00:00", cctz::format(RFC3339_full, tp, utc));
    // 1 minute beyond max should fail.
    EXPECT_FALSE(parse(RFC3339_full, "1970-01-01T02:08:00+00:00", utc, &tp));

    // Min time_point<D> is 1969-12-31T21:52:00+00:00.
    EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T21:52:00+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::min());
    EXPECT_EQ("1969-12-31T21:52:00+00:00", cctz::format(RFC3339_full, tp, utc));
    // 1 minute below min should fail.
    EXPECT_FALSE(parse(RFC3339_full, "1969-12-31T21:51:00+00:00", utc, &tp));
  }
  {
    // Test with millisecond resolution using int8_t ([-128, 127] milliseconds).
    // This tests the case where Denom (1000) is larger than Rep max/min.
    using D = chrono::duration<std::int8_t, std::milli>;
    time_point<D> tp;

    // Max time_point<D> is 1970-01-01T00:00:00.127+00:00.
    EXPECT_TRUE(parse(RFC3339_full, "1970-01-01T00:00:00.127+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::max());
    // 1 millisecond beyond max should fail (tests sub > max check).
    EXPECT_FALSE(
        parse(RFC3339_full, "1970-01-01T00:00:00.128+00:00", utc, &tp));

    // Min time_point<D> is 1969-12-31T23:59:59.872+00:00.
    EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:59:59.872+00:00", utc, &tp));
    EXPECT_EQ(tp, time_point<D>::min());
    // 1 millisecond below min should fail (tests min with Denom > -min).
    EXPECT_FALSE(
        parse(RFC3339_full, "1969-12-31T23:59:59.871+00:00", utc, &tp));
  }
}
```

src/time_zone_format_test.cc
```diff
@@ -1628,52 +1628,144 @@ TEST(Parse, MaxRange) {
...
+  EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:59:59+00:00", utc, &dtp));
+  EXPECT_EQ(dtp.time_since_epoch().count(), -1.0);
+
+  using DD_milli = chrono::duration<double, std::milli>;
+  time_point<DD_milli> mdtp;
+
+  EXPECT_TRUE(parse(RFC3339_full, "1970-01-01T00:00:00.5+00:00", utc, &mdtp));
+  EXPECT_EQ(mdtp.time_since_epoch().count(), 500.0);
+  EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:59:59.5+00:00", utc, &mdtp));
+  EXPECT_EQ(mdtp.time_since_epoch().count(), -500.0);
```
devbww: These don't have anything to do with "TimePointOverflow", so I'd place them in a separate test.

It also seems like we should be testing fractional values and periods larger than 1 second (which would catch the `join_second()` issue mentioned above).

So, ...
```
TEST(Parse, TimePointFloatingPoint) {
  const time_zone utc = utc_time_zone();
  {
    using D = chrono::duration<double>;
    time_point<D> tp;

    EXPECT_TRUE(parse(RFC3339_full, "1970-01-01T00:00:59.25+00:00", utc, &tp));
    EXPECT_EQ(tp.time_since_epoch(), D{59.25});
    EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:59:00.75+00:00", utc, &tp));
    EXPECT_EQ(tp.time_since_epoch(), D{-59.25});
  } 
  { 
    using D = chrono::duration<double, std::milli>;
    time_point<D> tp;
    
    EXPECT_TRUE(
        parse(RFC3339_full, "1970-01-01T00:00:00.015625+00:00", utc, &tp));
    EXPECT_EQ(tp.time_since_epoch(), D{15.625});
    EXPECT_TRUE(
        parse(RFC3339_full, "1969-12-31T23:59:59.984375+00:00", utc, &tp));
    EXPECT_EQ(tp.time_since_epoch(), D{-15.625});
  }
  {
    using D = chrono::duration<double, std::ratio<60>>;
    time_point<D> tp;
    
    EXPECT_TRUE(parse(RFC3339_full, "1970-01-01T00:01:15+00:00", utc, &tp));
    EXPECT_EQ(tp.time_since_epoch(), D{1.25});
    EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:58:45+00:00", utc, &tp));
    EXPECT_EQ(tp.time_since_epoch(), D{-1.25});
  } 
}
```

include/cctz/time_zone.h
```diff
@@ -423,7 +472,7 @@ bool join_seconds(
     count -= 1;
   }
   if (count > (std::numeric_limits<Rep>::max)()) return false;
-  if (count < (std::numeric_limits<Rep>::min)()) return false;
+  if (count < (std::numeric_limits<Rep>::lowest)()) return false;
   *tpp = time_point<D>() + D{static_cast<Rep>(count)};
```
devbww, line 480: If I was writing this again, I'd say ...
```
  *tpp = std::chrono::time_point_cast<D>(sec);
```
Perhaps now is a good time to change it.

include/cctz/time_zone.h
```diff
@@ -435,7 +484,7 @@ bool join_seconds(
   using D = std::chrono::duration<Rep, std::ratio<1, 1>>;
   auto count = sec.time_since_epoch().count();
   if (count > (std::numeric_limits<Rep>::max)()) return false;
-  if (count < (std::numeric_limits<Rep>::min)()) return false;
+  if (count < (std::numeric_limits<Rep>::lowest)()) return false;
   *tpp = time_point<D>() + D{static_cast<Rep>(count)};
```
devbww, line 493: Ditto.

include/cctz/time_zone.h
devbww: Here is the idea to replace the existing `cctz::parse()`:
```
template <typename Rep, typename Period>
bool parse(const std::string& fmt, const std::string& input,
           const time_zone& tz,
           time_point<std::chrono::duration<Rep, Period>>* tpp) {
  time_point<seconds> sec;
  detail::femtoseconds fs;
  if (!detail::parse(fmt, input, tz, &sec, &fs)) return false;
  time_point<std::chrono::duration<Rep, typename Period::type>> tp;
  if (!detail::join_seconds(sec, fs, &tp)) return false;
  *tpp = tp;
  return true;
}
```
And a test that previously would not have compiled could be something like:
```
TEST(Parse, RatioNonLowestTerms) {
  const time_zone utc = utc_time_zone();
  using D = chrono::duration<std::int64_t, std::ratio<2, 2>>;
  time_point<D> tp;

  EXPECT_TRUE(parse(RFC3339_full, "2026-08-17T11:42:16-04:00", utc, &tp));
  ExpectTime(tp, utc, 2026, 8, 17, 15, 42, 16, 0, false, "UTC");
}
```

include/cctz/time_zone.h
derekmauro: Done.

include/cctz/time_zone.h
derekmauro: Used enable_if as suggested below.

include/cctz/time_zone.h
derekmauro: Done.

include/cctz/time_zone.h
derekmauro: Done.

include/cctz/time_zone.h
derekmauro: Done.

include/cctz/time_zone.h
derekmauro: This comment sent me into a bit of a rabbit hole. In libc++, ratios are reduced by std::ratio_divide, so with the SI units, it will usually help. If you are using some strange ratios with a small GCD, you are going to have a bad time.

src/time_zone_format_test.cc
derekmauro: Done.

src/time_zone_format_test.cc
derekmauro: Done.

src/time_zone_format_test.cc
derekmauro: Done.

src/time_zone_format_test.cc
derekmauro: Done.

include/cctz/time_zone.h
derekmauro: This works below, but here std::chrono::time_point_cast truncates towards zero, which breaks a bunch of stuff.

include/cctz/time_zone.h
derekmauro: Done.

include/cctz/time_zone.h
devbww: I didn't see any breakages (in the existing tests at least).  Can you give more details?

If rounding behavior is indeed an issue (and given that we can't use `std::chrono::floor()`/`ceil()`/`round()` until we require C++17), perhaps we can take explicit action after `time_point_cast<D>()` to correct for its truncating behavior, like we currently do in `split_seconds()`.

include/cctz/time_zone.h
devbww: In the interests of getting rid of as many `.count()`s as possible (i.e., leaving things in the `chrono` domain), let's change this to ...
```
  if (sub < D::zero()) {
```
while we're here.

include/cctz/time_zone.h
```diff
@@ -364,13 +370,17 @@ inline std::string format(const std::string& fmt, const time_point<D>& tp,
 //   if (cctz::parse("%Y-%m-%d", "2015-10-09", tz, &tp)) {
 //     ...
 //   }
-template <typename D>
+template <typename Rep, typename Period>
 inline bool parse(const std::string& fmt, const std::string& input,
```
devbww: Should we get rid of the `inline` now?

include/cctz/time_zone.h
devbww: OK, thanks.  Sorry for the rabbit hole.  At the very least we can say it is not our problem.

include/cctz/time_zone.h
```diff
@@ -266,15 +267,20 @@ std::string format(const std::string&, const time_point<seconds>&,
 bool parse(const std::string&, const std::string&, const time_zone&,
            time_point<seconds>*, femtoseconds*, std::string* err = nullptr);
 template <typename Rep, std::intmax_t Denom>
-bool join_seconds(
+typename std::enable_if<std::is_integral<Rep>::value, bool>::type join_seconds(
```
devbww: Optional: It is a shame that the function name gets hidden at the end of these long lines (here and below).  Should we consider forcing line breaks, or are `namespace detail` forward declarations not worth the special treatment?  Does the same answer apply to the function definitions as well?

include/cctz/time_zone.h
derekmauro: I added tests to demonstrate the issue.

include/cctz/time_zone.h
derekmauro: Done.

include/cctz/time_zone.h
derekmauro: The rabbit hole was for my own curiosity. There are some things that I just have to know how they work, so I consider it a positive.

Manually formatted the code here.

include/cctz/time_zone.h
devbww: > I added tests to demonstrate the issue.

Thanks!  With those, and thinking more about `time_point_cast`'s behavior, I think the best thing to do is withdraw my suggestion about using it, and revert the integral, `std::ratio<1, 1>` version of `join_seconds()` back to use ...
```
  *tpp = time_point<D>() + D{static_cast<Rep>(count)};
```
Indeed, that also makes it much clearer that all the checks we made on the bounds of `count` have a direct consequence.

And considering a future with C++17 and `std::chrono::floor()` for a `time_point`, my last suggestion is to simplify `split_seconds()` to ...
```
template <typename D>
std::pair<time_point<seconds>, D> split_seconds(const time_point<D>& tp) {
  auto sec = std::chrono::time_point_cast<seconds>(tp);
  if (sec > tp) sec -= seconds{1};  // TODO(C++17): use std::chrono::floor
  return {sec, std::chrono::duration_cast<D>(tp - sec)};
}
```
It's clearer, and prepares us for that future.

derekmauro: Thanks, both changes made.

Speaking of C++17, how should we handle that? Do one more release and then declare C++17 as the new floor?

devbww: > Speaking of C++17, how should we handle that? Do one more release and then declare C++17 as the new floor?

That sounds like a good plan, although, `std::chrono::floor()` notwithstanding, I'm certainly in no hurry to do that.  I guess with C++17 there would be opportunities to replace `const std::string&` with `std::string_view` in some interfaces.  What else might be useful?

devbww (approved):
LGTM.  Thanks for taking this on, and for entertaining my suggestions.

include/cctz/time_zone.h
```diff
@@ -364,13 +373,17 @@ inline std::string format(const std::string& fmt, const time_point<D>& tp,
...
-                  const time_zone& tz, time_point<D>* tpp) {
+template <typename Rep, typename Period>
+bool parse(const std::string& fmt, const std::string& input,
+           const time_zone& tz,
+           time_point<std::chrono::duration<Rep, Period>>* tpp) {
   time_point<seconds> sec;
   detail::femtoseconds fs;
-  return detail::parse(fmt, input, tz, &sec, &fs) &&
-         detail::join_seconds(sec, fs, tpp);
+  if (!detail::parse(fmt, input, tz, &sec, &fs)) return false;
```
dinord, line 382: Would you mind adding putting an empty line after each `return false` to create groups of lines of code that initialize each variable?

src/time_zone_format_test.cc
```diff
@@ -787,10 +788,14 @@ TEST(Parse, TimePointResolution) {
   time_point<chrono::minutes> tp_m;
   EXPECT_TRUE(parse(kFmt, "03:04:05", utc, &tp_m));
   EXPECT_EQ("03:04:00", cctz::format(kFmt, tp_m, utc));
+  EXPECT_TRUE(parse(RFC3339_full, "1969-12-31T23:59:05+00:00", utc, &tp_m));
```
dinord, line 793: Consider delineating this with empty lines and adding a brief comment that explains what is being tested. I can't confidently tell without reverse-engineering.

Same for the block on line 797.

dinord (approved):

include/cctz/time_zone.h
derekmauro: Done.

src/time_zone_format_test.cc
derekmauro: Done.

pkumar0508:
(Kudos to Dino for 🏎️ , looks like I could brush up my reading speed with std::chrono APIs)

PR #356, opened 2026-07-21
avoid signed overflow computing start offset in AndroidZoneInfoSource

rajath201:
AndroidZoneInfoSource::Open decodes each index entry's absolute data offset as `data_offset + Decode32(ebuf + 40)`, where both operands come from a loaded, untrusted tzdata database. `std::int_fast32_t` is 32 bits on Bionic, macOS, and Windows, so a crafted index with `data_offset` and an entry offset near `INT32_MAX` overflows the addition before the `start < 0` guard ever runs (UBSan flags `2147483647 + 2147483646`).

Do the sum in `int_fast64_t` before the range check, the same way the TZif reader in `Load()` guards its decoded values. Real tzdata offsets stay well under `INT32_MAX`, so valid zones are unaffected; only the hostile case changes, and it now fails the bounds check cleanly instead of relying on undefined behavior.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -536,7 +536,8 @@ std::unique_ptr<ZoneInfoSource> AndroidZoneInfoSource::Open(
       const std::int_fast64_t start =
           static_cast<std::int_fast64_t>(data_offset) + Decode32(ebuf + 40);
       const std::int_fast32_t length = Decode32(ebuf + 44);
-      if (start < 0 || length < 0) break;
+      if (start < 0 || start > std::numeric_limits<long>::max() || length < 0)
+        break;  // fseek() takes a long
       ebuf[40] = '\0';  // ensure zone name is NUL terminated
       if (strcmp(name.c_str() + pos, ebuf) == 0) {
         if (fseek(fp.get(), static_cast<long>(start), SEEK_SET) != 0) break;
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -47,6 +47,7 @@
 #include <cstring>
 #include <fstream>
 #include <functional>
+#include <limits>
 #include <memory>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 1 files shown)


src/time_zone_info.cc
devbww, line 545: Should we then also be checking that the newly-widened `start` fits into a `long`?

rajath201: Good point, done. `long` is still 32 bits on LLP64 targets, so the cast at the `fseek()` could truncate a large `start`; the guard now also rejects entries with `start > std::numeric_limits<long>::max()`. Where `long` is 64 bits the check folds away, and gcc/clang stay quiet about it under the CI's -Wall/-Wextra.

src/time_zone_info.cc
```diff
@@ -533,9 +533,11 @@ std::unique_ptr<ZoneInfoSource> AndroidZoneInfoSource::Open(
     if (zonecnt * sizeof(ebuf) != index_size) continue;
     for (std::size_t i = 0; i != zonecnt; ++i) {
       if (fread(ebuf, 1, sizeof(ebuf), fp.get()) != sizeof(ebuf)) break;
-      const std::int_fast32_t start = data_offset + Decode32(ebuf + 40);
+      const std::int_fast64_t start =
+          static_cast<std::int_fast64_t>(data_offset) + Decode32(ebuf + 40);
       const std::int_fast32_t length = Decode32(ebuf + 44);
-      if (start < 0 || length < 0) break;
+      if (start < 0 || start > std::numeric_limits<long>::max() || length < 0)
```
devbww: It looks like somewhere along the way (2017), we failed to add `#include <limits>` to this file, so we might as well do it now.

src/time_zone_info.cc
```diff
@@ -533,9 +533,11 @@ std::unique_ptr<ZoneInfoSource> AndroidZoneInfoSource::Open(
...
+        break;  // fseek() takes a long
```
devbww: This new comment only applies to one of the conditionals, so I'd either just remove the comment, or separate the "fseek() takes a long" conditional into its own statement.

src/time_zone_info.cc
```diff
@@ -533,9 +533,11 @@ std::unique_ptr<ZoneInfoSource> AndroidZoneInfoSource::Open(
     if (zonecnt * sizeof(ebuf) != index_size) continue;
     for (std::size_t i = 0; i != zonecnt; ++i) {
       if (fread(ebuf, 1, sizeof(ebuf), fp.get()) != sizeof(ebuf)) break;
-      const std::int_fast32_t start = data_offset + Decode32(ebuf + 40);
+      const std::int_fast64_t start =
+          static_cast<std::int_fast64_t>(data_offset) + Decode32(ebuf + 40);
```
devbww: Following an idea from #360, we're better off using brace initialization instead of `static_cast` for a widening integer conversion.  That is, `std::int_fast64_t{data_offset}`.

src/time_zone_info.cc
rajath201: Done, added to the include block.

src/time_zone_info.cc
rajath201: Went with the split. The long-range check is now its own statement with the comment on it, and the original start/length check reads the same as before the PR.

src/time_zone_info.cc
rajath201: Done, switched to std::int_fast64_t{data_offset}. Nice property that it'd fail to compile if the conversion ever became narrowing.

devbww (approved):
Looks good to me.  Thanks.

We'll have to wait for a Google owner to initiate the tests.

mkruskal-google (approved):

PR #357, opened 2026-07-21
Reject extended zones whose transitions remain in negative unix time

derekmauro:
When a TZif file has explicit transitions ending before the epoch and
includes a POSIX DST footer string, ExtendTransitions() sets extended_ = true
and generates 401 years of transitions. Load() previously checked
`if (last.unix_time < 0)` unconditionally and appended a static sentinel
transition at unix_time = 2147483647 (2038).
  
Appending a static transition to an extended zone table is incompatible with
the 400-year cycle shifting used for future lookups.
  
This change:
   - Updates TimeZoneInfo::Load() to reject zoneinfo files if 401 years of
     extended transitions fail to reach the non-negative unix time space
     (`if (extended_) return false;`). All valid zoneinfo files easily meet
     this requirement.
   - Avoids appending a static no-op transition when extended_ is true,
     preserving the generated 400-year cycle transition table.
   - Guarantees that for any successfully loaded extended zone, `last.unix_time`
     and `last_year_` are non-negative, ensuring that 400-year cycle calculations
     in BreakTime() and MakeTime() operate safely within bounds.
   - Adds unit test coverage in `time_zone_lookup_test.cc` to verify lookups for
     valid zones with transitions before the epoch, and to confirm rejection of
     zones whose extended transitions fail to reach non-negative unix time.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -800,17 +800,13 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
   // time line (the first half is handled above) so that the signed
   // difference between a civil_second and the civil_second of its
   // previous transition is always representable, without overflow.
-  // Note: For zones extended with a future POSIX spec (extended_ == true),
-  // future times are handled by 400-year cycle shifting. Appending a static
-  // transition would corrupt the extended transition table.
-  if (!extended_) {
-    const Transition& last(transitions_.back());
-    if (last.unix_time < 0) {
-      const std::uint_fast8_t type_index = last.type_index;
-      Transition& tr(*transitions_.emplace(transitions_.end()));
-      tr.unix_time = 2147483647;  // 2038-01-19T03:14:07+00:00
-      tr.type_index = type_index;
-    }
+  const Transition& last(transitions_.back());
+  if (last.unix_time < 0) {
+    if (extended_) return false;
+    const std::uint_fast8_t type_index = last.type_index;
+    Transition& tr(*transitions_.emplace(transitions_.end()));
+    tr.unix_time = 2147483647;  // 2038-01-19T03:14:07+00:00
+    tr.type_index = type_index;
   }
 
   // Compute the local civil time for each transition and the preceding
@@ -929,17 +925,9 @@ time_zone::absolute_lookup TimeZoneInfo::BreakTime(
     // future_spec_, shift back to a supported year using the 400-year
     // cycle of calendaric equivalence and then compensate accordingly.
     if (extended_) {
-      // Use unsigned 64-bit subtraction to avoid signed overflow when
-      // unix_time > 0 and the last transition's unix_time is negative.
-      const std::uint_fast64_t diff =
-          static_cast<std::uint_fast64_t>(unix_time) -
-          static_cast<std::uint_fast64_t>(transitions_[timecnt - 1].unix_time);
-      year_t shift = static_cast<year_t>(diff / kSecsPer400Years + 1);
-      // Clamp shift to prevent shift * kSecsPer400Years from overflowing
-      // seconds when querying far-future timestamps (e.g. seconds::max()).
-      if (shift > seconds::max().count() / kSecsPer400Years) {
-        shift = seconds::max().count() / kSecsPer400Years;
-      }
+      const std::int_fast64_t diff =
+          unix_time - transitions_[timecnt - 1].unix_time;
+      const year_t shift = diff / kSecsPer400Years + 1;
       const auto d = seconds(shift * kSecsPer400Years);
       time_zone::absolute_lookup al = BreakTime(tp - d);
       al.cs = YearShift(al.cs, shift * 400);
@@ -1012,19 +1000,8 @@ time_zone::civil_lookup TimeZoneInfo::MakeTime(const civil_second& cs) const {
       // future_spec_, shift back to a supported year using the 400-year
       // cycle of calendaric equivalence and then compensate accordingly.
       if (extended_ && cs.year() > last_year_) {
-        // Use unsigned 64-bit arithmetic to safely compute the year difference
-        // and avoid signed overflow when last_year_ is negative.
-        const std::uint_fast64_t diff =
-            static_cast<std::uint_fast64_t>(cs.year()) -
-            static_cast<std::uint_fast64_t>(last_year_ + 1);
-        const year_t shift = static_cast<year_t>(diff / 400 + 1);
-        // Map cs.year() to the 400-year cycle in (last_year_ - 400, last_year_]
-        // without intermediate multiplications that could overflow year_t.
-        const year_t shifted_year =
-            last_year_ - 399 + static_cast<year_t>(diff % 400);
-        const civil_second shifted_cs(shifted_year, cs.month(), cs.day(),
-                                      cs.hour(), cs.minute(), cs.second());
-        return TimeLocal(shifted_cs, shift);
+        const year_t shift = (cs.year() - last_year_ - 1) / 400 + 1;
+        return TimeLocal(YearShift(cs, shift * -400), shift);
       }
       const TransitionType& tt(transition_types_[tr->type_index]);
       if (cs > tt.civil_max) return MakeUnique(time_point<seconds>::max());
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -800,12 +800,17 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
   // time line (the first half is handled above) so that the signed
   // difference between a civil_second and the civil_second of its
   // previous transition is always representable, without overflow.
-  const Transition& last(transitions_.back());
-  if (last.unix_time < 0) {
-    const std::uint_fast8_t type_index = last.type_index;
-    Transition& tr(*transitions_.emplace(transitions_.end()));
-    tr.unix_time = 2147483647;  // 2038-01-19T03:14:07+00:00
-    tr.type_index = type_index;
+  // Note: For zones extended with a future POSIX spec (extended_ == true),
+  // future times are handled by 400-year cycle shifting. Appending a static
+  // transition would corrupt the extended transition table.
+  if (!extended_) {
+    const Transition& last(transitions_.back());
+    if (last.unix_time < 0) {
+      const std::uint_fast8_t type_index = last.type_index;
+      Transition& tr(*transitions_.emplace(transitions_.end()));
+      tr.unix_time = 2147483647;  // 2038-01-19T03:14:07+00:00
+      tr.type_index = type_index;
+    }
   }
 
   // Compute the local civil time for each transition and the preceding
@@ -924,9 +929,17 @@ time_zone::absolute_lookup TimeZoneInfo::BreakTime(
     // future_spec_, shift back to a supported year using the 400-year
     // cycle of calendaric equivalence and then compensate accordingly.
     if (extended_) {
-      const std::int_fast64_t diff =
-          unix_time - transitions_[timecnt - 1].unix_time;
-      const year_t shift = diff / kSecsPer400Years + 1;
+      // Use unsigned 64-bit subtraction to avoid signed overflow when
+      // unix_time > 0 and the last transition's unix_time is negative.
+      const std::uint_fast64_t diff =
+          static_cast<std::uint_fast64_t>(unix_time) -
+          static_cast<std::uint_fast64_t>(transitions_[timecnt - 1].unix_time);
+      year_t shift = static_cast<year_t>(diff / kSecsPer400Years + 1);
+      // Clamp shift to prevent shift * kSecsPer400Years from overflowing
+      // seconds when querying far-future timestamps (e.g. seconds::max()).
+      if (shift > seconds::max().count() / kSecsPer400Years) {
+        shift = seconds::max().count() / kSecsPer400Years;
+      }
       const auto d = seconds(shift * kSecsPer400Years);
       time_zone::absolute_lookup al = BreakTime(tp - d);
       al.cs = YearShift(al.cs, shift * 400);
@@ -999,8 +1012,19 @@ time_zone::civil_lookup TimeZoneInfo::MakeTime(const civil_second& cs) const {
       // future_spec_, shift back to a supported year using the 400-year
       // cycle of calendaric equivalence and then compensate accordingly.
       if (extended_ && cs.year() > last_year_) {
-        const year_t shift = (cs.year() - last_year_ - 1) / 400 + 1;
-        return TimeLocal(YearShift(cs, shift * -400), shift);
+        // Use unsigned 64-bit arithmetic to safely compute the year difference
+        // and avoid signed overflow when last_year_ is negative.
+        const std::uint_fast64_t diff =
+            static_cast<std::uint_fast64_t>(cs.year()) -
+            static_cast<std::uint_fast64_t>(last_year_ + 1);
+        const year_t shift = static_cast<year_t>(diff / 400 + 1);
+        // Map cs.year() to the 400-year cycle in (last_year_ - 400, last_year_]
+        // without intermediate multiplications that could overflow year_t.
+        const year_t shifted_year =
+            last_year_ - 399 + static_cast<year_t>(diff % 400);
+        const civil_second shifted_cs(shifted_year, cs.month(), cs.day(),
+                                      cs.hour(), cs.minute(), cs.second());
+        return TimeLocal(shifted_cs, shift);
       }
       const TransitionType& tt(transition_types_[tr->type_index]);
       if (cs > tt.civil_max) return MakeUnique(time_point<seconds>::max());
--- a/src/time_zone_lookup_test.cc
+++ b/src/time_zone_lookup_test.cc
@@ -950,8 +950,11 @@ class StringZoneInfoSource : public ZoneInfoSource {
 };
 
 // Constructs a minimal valid TZif2 string with a single 64-bit transition
-// at the given unix_time and a future POSIX DST footer ("EST5EDT,M3.2.0,M11.1.0").
-std::string MakeNegativeEpochTzif(std::int_fast64_t unix_time) {
+// at the given transition time and a future POSIX rule.
+std::string MakeExtendedTzif(std::int_fast64_t unix_time,
+                             std::int_fast32_t utc_offset,
+                             const std::string& abbr,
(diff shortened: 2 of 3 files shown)


src/time_zone_info.cc
```diff
@@ -839,12 +800,17 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
...
-  const Transition& last(transitions_.back());
-  if (last.unix_time < 0) {
-    const std::uint_fast8_t type_index = last.type_index;
-    Transition& tr(*transitions_.emplace(transitions_.end()));
-    tr.unix_time = 2147483647;  // 2038-01-19T03:14:07+00:00
-    tr.type_index = type_index;
+  // Note: For zones extended with a future POSIX spec (extended_ == true),
+  // future times are handled by 400-year cycle shifting. Appending a static
+  // transition would corrupt the extended transition table.
+  if (!extended_) {
```
devbww: Not having a transition in the second half of the "unix_time" space breaks the guarantee that the difference between a transition's civil second and that of its predecessor is always representable.

I agree that there is an issue here though.  However, the real problem is that extending the transitions with shifts in and out of daylight time is fundamentally incompatible with also later appending a "no-op" transition.

So, I instead suggest that we require that the 401 years of extended daylight-time transitions reach the second (positive) half of the "unix_time" space, removing the need to add the "no-op" transition.  Any real zoneinfo will meet this requirement.  (This is similar to the recent change to reject zoneinfo with extreme transition times.)

That is, just ...
```
@@ -841,6 +841,7 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
   // previous transition is always representable, without overflow.
   const Transition& last(transitions_.back());
   if (last.unix_time < 0) {
+    if (extended_) return false;
     const std::uint_fast8_t type_index = last.type_index;
     Transition& tr(*transitions_.emplace(transitions_.end()));
     tr.unix_time = 2147483647;  // 2038-01-19T03:14:07+00:00
```

src/time_zone_info.cc
```diff
@@ -963,9 +929,17 @@ time_zone::absolute_lookup TimeZoneInfo::BreakTime(
     // future_spec_, shift back to a supported year using the 400-year
     // cycle of calendaric equivalence and then compensate accordingly.
     if (extended_) {
-      const std::int_fast64_t diff =
-          unix_time - transitions_[timecnt - 1].unix_time;
-      const year_t shift = diff / kSecsPer400Years + 1;
+      // Use unsigned 64-bit subtraction to avoid signed overflow when
+      // unix_time > 0 and the last transition's unix_time is negative.
```
devbww: With the above edit we now know, in the `extended_` case, that the last transition's `unix_time` is non-negative.  Therefore, there is no need for any change here.

src/time_zone_info.cc
```diff
@@ -1038,8 +1012,19 @@ time_zone::civil_lookup TimeZoneInfo::MakeTime(const civil_second& cs) const {
       // future_spec_, shift back to a supported year using the 400-year
       // cycle of calendaric equivalence and then compensate accordingly.
       if (extended_ && cs.year() > last_year_) {
-        const year_t shift = (cs.year() - last_year_ - 1) / 400 + 1;
-        return TimeLocal(YearShift(cs, shift * -400), shift);
+        // Use unsigned 64-bit arithmetic to safely compute the year difference
+        // and avoid signed overflow when last_year_ is negative.
```
devbww: With the above edit we now know, in the `extended_` case, that `last_year_` is non-negative. Therefore, there is no need for any change here.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +910,123 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+
+// Tests loading a TZif file whose explicit transitions end in negative epoch
+// with a POSIX DST footer string. Verifies that querying far-future time_points
+// does not cause signed integer overflow, and that lookups for modern dates use
+// the correct POSIX DST transition rules.
+TEST(TimeZoneEdgeCase, NegativeEpochExtended) {
+  auto prev_factory = cctz_extension::zone_info_source_factory;
+  cctz_extension::zone_info_source_factory = NegativeEpochTestFactory;
+  time_zone tz;
+  ASSERT_TRUE(load_time_zone("test:negative_epoch_dst", &tz));
```
devbww: With the above edit this assertion will now fail.

I'm not sure if that means you want to keep a modified version of the test case.  But if you do, I'll add some additional test comments.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +910,123 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+ private:
+  std::string data_;
+  std::size_t offset_;
+};
+
+// Constructs a minimal valid TZif2 string with a single negative-epoch 64-bit
+// transition and a future POSIX DST footer ("EST5EDT,M3.2.0,M11.1.0").
+std::string MakeNegativeEpochTzif() {
+  std::string s;
+  auto append32 = [&s](std::uint32_t v) {
```
devbww: In CCTZ I have avoided the exact-size integer types, which are not even guaranteed to exist (although I do note that some uses have made their way into the format, fuzz, and win tests).  It seems like `std::uint_fast32_t` and `std::uint_fast64_t` will serve well here.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +910,123 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+    }
+  };
+  auto append64 = [&s](std::uint64_t v) {
+    for (int i = 7; i >= 0; --i) {
+      s.push_back(static_cast<char>((v >> (i * 8)) & 0xff));
+    }
+  };
+
+  // 32-bit header
+  s.append("TZif2", 5);
```
devbww: I'd split the `tzh_magic` and `tzh_version` components, and make use of `TZ_MAGIC`.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +910,123 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+  append32(0);         // tzh_ttisutcnt
+  append32(0);         // tzh_ttisstdcnt
+  append32(0);         // tzh_leapcnt
+  append32(1);         // tzh_timecnt (1 64-bit transition)
+  append32(1);         // tzh_typecnt (1 ttinfo record)
+  append32(4);         // tzh_charcnt (4 bytes for "EST\0")
+
+  // 64-bit data block
+  append64(static_cast<std::uint64_t>(
+      std::int64_t{-100000000000}));             // unix_time (year -1201)
```
devbww: This (and above/below) looks like it assumes twos-complement representation of signed values, which I've tried to avoid doing.  See the inverse implementations of `Decode32()` and `Decode64()` for example.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +910,123 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+  cctz_extension::zone_info_source_factory = prev_factory;
+
+  // Far-future lookup must not trigger signed integer overflow.
+  // A sanitizer like UBSAN may flag the following line on overflow.
+  static_cast<void>(tz.lookup(time_point<seconds>::max()));
+
+  // Summer (July) must use the POSIX DST rule (EDT, UTC-4).
+  auto tp_july = convert(civil_second(2026, 7, 20, 12, 0, 0), tz);
+  ExpectTime(tp_july, tz, 2026, 7, 20, 12, 0, 0, -4 * 3600, true, "EDT");
+  EXPECT_EQ("EDT", std::string(tz.lookup(tp_july).abbr));
```
devbww: Given that these are both "C strings", I'd probably reach for `EXPECT_STREQ()` instead of creating a `std::string`.

src/time_zone_info.cc
derekmauro: Done.

src/time_zone_lookup_test.cc
derekmauro: Done.

src/time_zone_lookup_test.cc
derekmauro: Done.

src/time_zone_info.cc
derekmauro: Done.

src/time_zone_info.cc
derekmauro: Done.

src/time_zone_lookup_test.cc
derekmauro: Done.

src/time_zone_lookup_test.cc
derekmauro: Seems like twos-complement is only required after C++20. Done.

src/time_zone_lookup_test.cc
derekmauro: Done.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+  }
+
+ private:
+  std::string data_;
+  std::size_t offset_;
+};
+
+// Constructs a minimal valid TZif2 string with a single 64-bit transition
+// at the given unix_time and a future POSIX DST footer ("EST5EDT,M3.2.0,M11.1.0").
+std::string MakeNegativeEpochTzif(std::int_fast64_t unix_time) {
```
devbww: I'm not sure what _"negative epoch"_ really means, but it seems like this function doesn't care about the sign of `unix_time` in any case.  So, I'd rename it.

I would probably also complete the abstraction by taking the UTC offset and abbreviation of the transition, as well as the future rule, as parameters.  That should make the comments on the test zone names much clearer too.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+      uv = static_cast<std::uint_fast64_t>(v + s64max + 1) + s64maxU + 1;
+    }
+    for (int i = 7; i >= 0; --i) {
+      s.push_back(static_cast<char>((uv >> (i * 8)) & 0xff));
+    }
+  };
+
+  // 32-bit header
+  s.append(TZ_MAGIC, 4);
+  s.push_back('2');
```
devbww: Given that we comment every other data element, I'd add `// tzh_version` here (and below).

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+  append32(0);         // tzh_ttisstdcnt
+  append32(0);         // tzh_leapcnt
+  append32(0);         // tzh_timecnt (0 32-bit transitions)
+  append32(1);         // tzh_typecnt (1 ttinfo record)
+  append32(4);         // tzh_charcnt (4 bytes for "EST\0")
+
+  // 32-bit data block
+  append32(-18000);      // tt_utoff (-5 hours)
+  s.push_back(0);        // tt_isdst (standard time)
+  s.push_back(0);        // tt_desigidx ("EST\0")
```
devbww: I'd remove the `\0` from the comment here (and below).  It's not part of the abbreviation.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+  s.append("EST\0", 4);  // string table
```
devbww: Perhaps `s/string/abbreviation/` here (and below).

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+
+  // POSIX footer
+  s.append("\nEST5EDT,M3.2.0,M11.1.0\n");
+  return s;
+}
+
+std::unique_ptr<ZoneInfoSource> NegativeEpochTestFactory(
```
devbww: Perhaps "before epoch" is better than "negative epoch", here and elsewhere.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+  s.append(15, '\0');  // tzh_reserved
+  append32(0);         // tzh_ttisutcnt
+  append32(0);         // tzh_ttisstdcnt
+  append32(0);         // tzh_leapcnt
+  append32(1);         // tzh_timecnt (1 64-bit transition)
+  append32(1);         // tzh_typecnt (1 ttinfo record)
+  append32(4);         // tzh_charcnt (4 bytes for "EST\0")
+
+  // 64-bit data block
+  append64(unix_time);   // unix_time
```
devbww: Two `unix_time`s looks redundant.  Perhaps the comment should be "transition time".

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+        fallback) {
+  if (name == "test:negative_epoch_dst") {
+    // 1900-01-01T00:00:00Z (-2208988800) is in negative epoch, but extending
+    // 401 years reaches 2301 AD in the positive unix_time space.
+    return std::unique_ptr<ZoneInfoSource>(
+        new StringZoneInfoSource(MakeNegativeEpochTzif(-2208988800LL)));
+  }
+  if (name == "test:negative_epoch_dst_too_far") {
+    // Year -1201 (-100000000000) is so far in the negative epoch that extending
+    // 401 years still leaves the last transition before 1970 (<0).
```
devbww: `s/before 1970 (<0)/in the negative unix_time space/` to match above.

I'd probably also just say "unix time" instead of `unix_time` (here and immediately above) as there is no so-named variable to refer to.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+    return std::unique_ptr<ZoneInfoSource>(
+        new StringZoneInfoSource(MakeNegativeEpochTzif(-100000000000LL)));
+  }
+  return fallback(name);
+}
+
+// Tests loading a TZif file whose explicit transitions end in negative epoch
+// with a POSIX DST footer string.
+TEST(TimeZoneEdgeCase, NegativeEpochExtended) {
```
devbww: Again, "negative epoch" sounds strange.  The epoch (an instant) to which we refer is defined to be zero.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+  EXPECT_STREQ("EDT", tz.lookup(tp_july).abbr);
+
+  // Winter (January) must use EST (UTC-5).
+  auto tp_jan = convert(civil_second(2026, 1, 20, 12, 0, 0), tz);
+  ExpectTime(tp_jan, tz, 2026, 1, 20, 12, 0, 0, -5 * 3600, false, "EST");
+  EXPECT_STREQ("EST", tz.lookup(tp_jan).abbr);
+
+  // Extended zones that do not reach non-negative unix_time are rejected.
+  EXPECT_FALSE(load_time_zone("test:negative_epoch_dst_too_far", &tz));
+  cctz_extension::zone_info_source_factory = prev_factory;
```
devbww, line 1072: I'd separate this cleanup line from the previous expectation with a blank line (and I'd probably do the same thing after the two setup lines).

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,145 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+
+// Tests loading a TZif file whose explicit transitions end in negative epoch
+// with a POSIX DST footer string.
+TEST(TimeZoneEdgeCase, NegativeEpochExtended) {
+  auto prev_factory = cctz_extension::zone_info_source_factory;
+  cctz_extension::zone_info_source_factory = NegativeEpochTestFactory;
+  time_zone tz;
+  ASSERT_TRUE(load_time_zone("test:negative_epoch_dst", &tz));
+
+  // Summer (July) must use the POSIX DST rule (EDT, UTC-4).
```
devbww: The seasons (here and below) are strictly immaterial, so I'd remove their mentions.  What matters is how the civil time compares with the future rule.

devbww:
PS:  It looks like the _"This change"_ part of the PR description could now do with some re-wording.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,156 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
 
+// A ZoneInfoSource implementation backed by an in-memory string buffer.
+class StringZoneInfoSource : public ZoneInfoSource {
+ public:
+  explicit StringZoneInfoSource(std::string data)
+      : data_(std::move(data)), offset_(0) {}
+
+  std::size_t Read(void* ptr, std::size_t size) override {
+    if (offset_ >= data_.size())
+      return 0;
```
devbww: An invariant of this class is `offset_ <= data.size()`.

Given that the `offset_ == data.size()` case falls out naturally from the following code, I would suggest dropping this conditional altogether.

src/time_zone_lookup_test.cc
```diff
@@ -916,4 +922,156 @@ TEST(TimeZoneEdgeCase, UTC5DigitYear) {
...
+  append32(0);         // tzh_leapcnt
+  append32(0);         // tzh_timecnt (0 32-bit transitions)
+  append32(1);         // tzh_typecnt (1 ttinfo record)
+  append32(static_cast<std::int_fast32_t>(charcnt));  // tzh_charcnt
+
+  // 32-bit data block
+  append32(utc_offset);               // tt_utoff
+  s.push_back(0);                     // tt_isdst (standard time)
+  s.push_back(0);                     // tt_desigidx
+  s.append(abbr.data(), charcnt);     // abbreviation table
```
devbww: Even though `.c_str()` and (the `const char*` returning) `.data()` are indistinguishable since C++11, I've been using the former as a clear indicator that we are really using the NUL.

But I'll leave it to you as to whether it is worth continuing that convention here (and below).

devbww (approved):

mkruskal-google (approved):

PR #360, opened 2026-07-23
Avoid signed overflow in AndroidZoneInfoSource::Open offset math

ravi0800:
AndroidZoneInfoSource::Open() computes a zone's file position as data_offset + Decode32(ebuf + 40), where both operands are int_fast32_t values pulled straight from the tzdata index. On a build where int_fast32_t is a 32-bit type (32-bit Android/ARM), two near-INT_MAX index values sum past the type's range, so the addition is signed overflow before the start < 0 guard on the next line ever runs. UBSan on a crafted tzdata blob reports "signed integer overflow: 76 + 2147483647 cannot be represented in type int" right at that add.

Widen the left operand to int_fast64_t so the sum is formed in a type that holds it, which leaves the existing start < 0 / length < 0 rejection meaningful and the later static_cast<long>(start) for fseek unchanged. Doing the widening at the add keeps the fix inside the decode where the operand widths are still in view.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -533,7 +533,8 @@ std::unique_ptr<ZoneInfoSource> AndroidZoneInfoSource::Open(
     if (zonecnt * sizeof(ebuf) != index_size) continue;
     for (std::size_t i = 0; i != zonecnt; ++i) {
       if (fread(ebuf, 1, sizeof(ebuf), fp.get()) != sizeof(ebuf)) break;
-      const std::int_fast32_t start = data_offset + Decode32(ebuf + 40);
+      const std::int_fast64_t start =
+          std::int_fast64_t{data_offset} + Decode32(ebuf + 40);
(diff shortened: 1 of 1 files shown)


src/time_zone_info.cc
```diff
@@ -533,7 +533,8 @@ std::unique_ptr<ZoneInfoSource> AndroidZoneInfoSource::Open(
     if (zonecnt * sizeof(ebuf) != index_size) continue;
     for (std::size_t i = 0; i != zonecnt; ++i) {
       if (fread(ebuf, 1, sizeof(ebuf), fp.get()) != sizeof(ebuf)) break;
-      const std::int_fast32_t start = data_offset + Decode32(ebuf + 40);
+      const std::int_fast64_t start =
+          std::int_fast64_t{data_offset} + Decode32(ebuf + 40);
```
devbww, line 537: This looks like it is being handled in #356.  Please take a look there and see if you have anything to add.  Otherwise we can close this PR.  Thanks.

PR #361, opened 2026-07-24
avoid signed overflow computing 400-year shift in BreakTime

rajath201:
BreakTime()'s extended-zone branch shifts a far-future time back through the 400-year cycle with `shift = diff / kSecsPer400Years + 1`, then forms `seconds(shift * kSecsPer400Years)` in signed int64. When a zone is extended by a POSIX DST footer and its last transition sits low in positive unix time (explicit transitions before ~1795, e.g. a 1700 transition that extends to 2101), a lookup near `time_point<seconds>::max()` drives shift past `seconds::max()/kSecsPer400Years` and that multiply overflows. UBSan trips at time_zone_info.cc:979 on `730692562 * 12622780800`.

Per review, this now extends the restriction from #357 instead of clamping in BreakTime(): zoneinfo data with a rule for future transitions must end its explicit transitions at or after the epoch, so the 400-year cycle starts in non-negative unix time and `(diff / kSecsPer400Years + 1) * kSecsPer400Years` is always representable. The #357 check in Load() becomes an assert, before-epoch extended zones are rejected outright, and the far-future test looks up the maximum time in a zone whose final transition is at the epoch, the largest permitted shift.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -975,7 +975,11 @@ time_zone::absolute_lookup TimeZoneInfo::BreakTime(
     if (extended_) {
       const std::int_fast64_t diff =
           unix_time - transitions_[timecnt - 1].unix_time;
-      const year_t shift = diff / kSecsPer400Years + 1;
+      // Bound the number of 400-year cycles so that shift * kSecsPer400Years
+      // stays representable (mirrors the clamp in TimeLocal()). A cycle spans
+      // exactly 400 years, so the recursive call folds in any residual shift.
+      const year_t shift = std::min<year_t>(
+          diff / kSecsPer400Years + 1, seconds::max().count() / kSecsPer400Years);
       const auto d = seconds(shift * kSecsPer400Years);
       time_zone::absolute_lookup al = BreakTime(tp - d);
       al.cs = YearShift(al.cs, shift * 400);
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -342,6 +342,14 @@ bool TimeZoneInfo::ExtendTransitions() {
     return EquivTransitions(transitions_.back().type_index, dst_ti);
   }
 
+  // We require that zoneinfo data with a rule for future transitions
+  // ends with a non-negative transition.  This removes the need to add
+  // any "second-half" transition to ensure differences between adjacent
+  // transitions are always representable, while also guaranteeing that
+  // the arithmetic used to shift between 400-year cycles never overflows.
+  // All valid zones easily meet this requirement.
+  if (transitions_.back().unix_time < 0) return false;
+
   // Extend the transitions for an additional 401 years using the future
   // specification. Years beyond those can be handled by mapping back to
   // a cycle-equivalent year within that range. Note that we need 401
@@ -850,7 +858,7 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
   // previous transition is always representable, without overflow.
   const Transition& last(transitions_.back());
   if (last.unix_time < 0) {
-    if (extended_) return false;
+    assert(!extended_);
     const std::uint_fast8_t type_index = last.type_index;
     Transition& tr(*transitions_.emplace(transitions_.end()));
     tr.unix_time = 2147483647;  // 2038-01-19T03:14:07+00:00
@@ -975,11 +983,7 @@ time_zone::absolute_lookup TimeZoneInfo::BreakTime(
     if (extended_) {
       const std::int_fast64_t diff =
           unix_time - transitions_[timecnt - 1].unix_time;
-      // Bound the number of 400-year cycles so that shift * kSecsPer400Years
-      // stays representable (mirrors the clamp in TimeLocal()). A cycle spans
-      // exactly 400 years, so the recursive call folds in any residual shift.
-      const year_t shift = std::min<year_t>(
-          diff / kSecsPer400Years + 1, seconds::max().count() / kSecsPer400Years);
+      const year_t shift = diff / kSecsPer400Years + 1;
       const auto d = seconds(shift * kSecsPer400Years);
       time_zone::absolute_lookup al = BreakTime(tp - d);
       al.cs = YearShift(al.cs, shift * 400);
--- a/src/time_zone_lookup_test.cc
+++ b/src/time_zone_lookup_test.cc
@@ -1044,6 +1044,14 @@ std::unique_ptr<ZoneInfoSource> ExtendedTestFactory(
         MakeExtendedTzif(-100000000000LL, -5 * 3600, "EST",
                          "EST5EDT,M3.2.0,M11.1.0")));
   }
+  if (name == "test:extended_dst_far_future") {
+    // 1700-01-01T00:00:00Z (-8520336000) extends 401 years to 2101, so the
+    // last transition sits low enough in positive unix time that a lookup at
+    // the maximum time needs a very large 400-year shift in BreakTime().
+    return std::unique_ptr<ZoneInfoSource>(new StringZoneInfoSource(
+        MakeExtendedTzif(-8520336000LL, -5 * 3600, "EST",
+                         "EST5EDT,M3.2.0,M11.1.0")));
+  }
   return fallback(name);
 }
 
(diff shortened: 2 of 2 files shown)


src/time_zone_info.cc
```diff
@@ -975,7 +975,11 @@ time_zone::absolute_lookup TimeZoneInfo::BreakTime(
...
       const std::int_fast64_t diff =
           unix_time - transitions_[timecnt - 1].unix_time;
-      const year_t shift = diff / kSecsPer400Years + 1;
+      // Bound the number of 400-year cycles so that shift * kSecsPer400Years
+      // stays representable (mirrors the clamp in TimeLocal()). A cycle spans
+      // exactly 400 years, so the recursive call folds in any residual shift.
+      const year_t shift = std::min<year_t>(
+          diff / kSecsPer400Years + 1, seconds::max().count() / kSecsPer400Years);
       const auto d = seconds(shift * kSecsPer400Years);
       time_zone::absolute_lookup al = BreakTime(tp - d);
```
devbww, line 988: Reducing by only `seconds::max().count() / kSecsPer400Years` cycles and handling the residual recursively is certainly valid.  But so would reducing by one cycle (albeit with unacceptable recursion depth).

So it seems to me that it would be clearer if we directly split `shift` into two cycle-sized pieces, N and 1, and handled them both in the same call.  That is, the `shift = ... `**`+ 1`** is the real problem.  Therefore, ...
```
      const year_t shift = diff / kSecsPer400Years;
      time_zone::absolute_lookup al = BreakTime(
          tp - seconds(shift * kSecsPer400Years) - seconds(kSecsPer400Years));
      al.cs = YearShift(al.cs, (shift + 1) * 400);
```

That said, I think we would instead be best off if we simply extended the restriction introduced in #357 such that the 400-year cycle has to start in non-negative territory rather than end there.  This would guarantee that `(diff / kSecsPer400Years + 1) * kSecsPer400Years` is representable.
```
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -342,6 +342,14 @@ bool TimeZoneInfo::ExtendTransitions() {
     return EquivTransitions(transitions_.back().type_index, dst_ti);
   }
 
+  // We require that zoneinfo data with a rule for future transitions
+  // ends with a non-negative transition.  This removes the need to add
+  // any "second-half" transition to ensure differences between adjacent
+  // transitions are always representable, while also guaranteeing that
+  // the arithmetic used to shift between 400-year cycles never overflows.
+  // All valid zones easily meet this requirement.
+  if (transitions_.back().unix_time < 0) return false;
+
   // Extend the transitions for an additional 401 years using the future
   // specification. Years beyond those can be handled by mapping back to
   // a cycle-equivalent year within that range. Note that we need 401
@@ -850,7 +858,7 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
   // previous transition is always representable, without overflow.
   const Transition& last(transitions_.back());
   if (last.unix_time < 0) {
-    if (extended_) return false;
+    assert(!extended_);
     const std::uint_fast8_t type_index = last.type_index;
     Transition& tr(*transitions_.emplace(transitions_.end()));
     tr.unix_time = 2147483647;  // 2038-01-19T03:14:07+00:00
```

src/time_zone_lookup_test.cc
```diff
@@ -1072,4 +1080,23 @@ TEST(TimeZoneEdgeCase, ExtendedBeforeEpoch) {
...
 }
 
+// Looking up the maximum time in an extended zone must fold back through the
+// 400-year cycle without overflowing when BreakTime() computes the shift.
+TEST(TimeZoneEdgeCase, ExtendedFarFuture) {
+  auto prev_factory = cctz_extension::zone_info_source_factory;
+  cctz_extension::zone_info_source_factory = ExtendedTestFactory;
+
+  time_zone tz;
+  ASSERT_TRUE(load_time_zone("test:extended_dst_far_future", &tz));
```
devbww: The suggested change breaks this assertion as the zone will no longer load.

The same will be true of the recently added _"test:extended_dst"_, so we'll have to adjust that case.

src/time_zone_lookup_test.cc
```diff
@@ -1072,4 +1080,23 @@ TEST(TimeZoneEdgeCase, ExtendedBeforeEpoch) {
...
+
+  // The maximum representable instant lands in December, which the future
+  // rule (EST5EDT,M3.2.0,M11.1.0) resolves to standard time.
+  const auto al = tz.lookup(time_point<cctz::seconds>::max());
+  EXPECT_EQ(-5 * 3600, al.offset);
+  EXPECT_FALSE(al.is_dst);
+  EXPECT_STREQ("EST", al.abbr);
```
devbww: If some form of this test survives, I would formulate it like all the similar existing ones by using `ExpectTime()`.
```
  auto tp_max = time_point<cctz::seconds>::max();
  ExpectTime(tp_max, tz, 292277026596, 12, 4, 10, 30, 7, -5 * 3600, false,
             "EST");
  EXPECT_STREQ("EST", tz.lookup(tp_max).abbr);
```

devbww (changes requested):

src/time_zone_info.cc
rajath201: Went with the restriction, it's cleaner than patching one caller. Applied your ExtendTransitions() diff as-is, reverted the BreakTime() change, and made the Load() check an assert. With the last explicit transition non-negative, the extension ends past the seconds::max() % kSecsPer400Years boundary, so the +1 cycle always stays representable.

src/time_zone_lookup_test.cc
rajath201: Right. ExtendedBeforeEpoch now expects both test:extended_dst and test:extended_dst_too_far to be rejected.

src/time_zone_lookup_test.cc
rajath201: Kept it. The zone's transition moves up to 1970-01-01T00:00:00Z so it still loads, which is now the earliest an extended zone may end and so still maximizes the shift, and the test uses your ExpectTime() formulation. Passes clean under UBSan.

src/time_zone_info.cc
rajath201: Agreed, rejecting these zones at load beats patching the lookup path, and it leaves BreakTime() alone. Applied your diff as-is and dropped the clamp. Lookup and format suites are clean under UBSan, including a max-time lookup with the final transition at the epoch, which is the largest shift the restriction permits.

src/time_zone_lookup_test.cc
rajath201: Right, both of those zones end before the epoch, so ExtendedBeforeEpoch now expects rejection for test:extended_dst as well as the too_far case.

src/time_zone_lookup_test.cc
rajath201: Done. Kept the test with the final transition moved to unix time 0, the earliest now permitted, so the tp_max lookup still exercises the largest possible shift, and formulated it with ExpectTime() as you suggested.

src/time_zone_lookup_test.cc
```diff
@@ -1031,43 +1031,58 @@ std::unique_ptr<ZoneInfoSource> ExtendedTestFactory(
...
   }
   if (name == "test:extended_dst_too_far") {
-    // Year -1201 (-100000000000) is so far before epoch that extending
-    // 401 years still leaves the last transition in the negative unix time space.
+    // Year -1201 (-100000000000) is so far before epoch that even extending
+    // 401 years would leave the last transition in the negative unix time space.
     return std::unique_ptr<ZoneInfoSource>(new StringZoneInfoSource(
         MakeExtendedTzif(-100000000000LL, -5 * 3600, "EST",
                          "EST5EDT,M3.2.0,M11.1.0")));
   }
```
devbww: I would suggest that we now only need one of these, whose single transition is at -1.

You might now also be able to choose better _"test:..."_ names for that "-1" zone and your new "0" one.  I would suggest that the zone name exactly match the TEST() case that uses it.

It doesn't seem like we need to add "LL" to the -1 and 0 constants either.

devbww (commented):
Looking good.  Thanks.

src/time_zone_lookup_test.cc
rajath201: Done. Collapsed the rejection cases into a single zone with its transition at -1, renamed both zones to match their TEST() cases (test:ExtendedBeforeEpoch, test:ExtendedFarFuture), and dropped the LL suffixes. Suite still passes clean under UBSan.

devbww (approved):
Excellent.  Thanks.

I'll pass off to the Google owners, @derekmauro and @mkruskal-google.

derekmauro (approved):

PR #362, opened 2026-07-28
reject unterminated abbreviations in TimeZoneInfo::Load

rajath201:
Load() validates every other decoded field but takes the abbreviation area on faith: tzfile.h describes it as "'\0'-terminated zone abbreviations" and abbr_index is checked against charcnt, yet nothing confirms the area itself ends in a NUL. Since LocalTime() hands out a bare `const char*` into `abbreviations_` as `absolute_lookup::abbr`, a zone carrying the three bytes "EST" with an `EST5EDT,M3.2.0,M11.1.0` footer loads fine and `lookup().abbr` comes back "ESTEDT", running on into the abbreviation GetTransitionType() appended behind it. Widening the footer to `EST5<XYZLONGABBREVIATION>` stretches that same lookup to 22 characters.

Reject it in Load(), alongside the existing charcnt and abbr_index checks. Keeping the guarantee on the one buffer every abbr pointer aliases beats asking each reader to bound its own string, and the other producers of that area (GetTransitionType(), ResetToBuiltinUTC()) already append the terminator themselves, so this only covers what comes off disk. All 485 zones under testdata/zoneinfo still load and convert unchanged.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -802,8 +802,9 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
   // Copy all the abbreviations. The area holds NUL-terminated strings, and
   // LocalTime() hands out a pointer into it, so the final abbreviation has
   // to be terminated within the area itself. Otherwise an abbreviation runs
-  // on into whatever ExtendTransitions() later appends.
-  if (hdr.charcnt == 0 || bp[hdr.charcnt - 1] != '\0')
+  // on into whatever ExtendTransitions() later appends. (hdr.charcnt != 0
+  // because every abbr_index was validated to be less than it.)
+  if (bp[hdr.charcnt - 1] != '\0')
     return false;
   abbreviations_.reserve(hdr.charcnt + 10);
   abbreviations_.assign(bp, hdr.charcnt);
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -799,7 +799,12 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
       default_transition_type_ = static_cast<std::uint_fast8_t>(index);
   }
 
-  // Copy all the abbreviations.
+  // Copy all the abbreviations. The area holds NUL-terminated strings, and
+  // LocalTime() hands out a pointer into it, so the final abbreviation has
+  // to be terminated within the area itself. Otherwise an abbreviation runs
+  // on into whatever ExtendTransitions() later appends.
+  if (hdr.charcnt == 0 || bp[hdr.charcnt - 1] != '\0')
+    return false;
   abbreviations_.reserve(hdr.charcnt + 10);
   abbreviations_.assign(bp, hdr.charcnt);
   bp += hdr.charcnt;
--- a/src/time_zone_lookup_test.cc
+++ b/src/time_zone_lookup_test.cc
@@ -1047,33 +1047,23 @@ std::unique_ptr<ZoneInfoSource> ExtendedTestFactory(
         MakeExtendedTzif(0, -5 * 3600, std::string{"EST", 4},
                          "EST5EDT,M3.2.0,M11.1.0")));
   }
-  if (name == "test:terminated_abbr") {
-    return std::unique_ptr<ZoneInfoSource>(new StringZoneInfoSource(
-        MakeExtendedTzif(0, -5 * 3600, std::string{"EST", 4},
-                         "EST5EDT,M3.2.0,M11.1.0")));
-  }
-  if (name == "test:unterminated_abbr") {
+  if (name == "test:UnterminatedAbbreviation") {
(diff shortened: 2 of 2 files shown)


src/time_zone_info.cc
```diff
@@ -791,7 +791,12 @@ bool TimeZoneInfo::Load(ZoneInfoSource* zip) {
       default_transition_type_ = static_cast<std::uint_fast8_t>(index);
   }
 
-  // Copy all the abbreviations.
+  // Copy all the abbreviations. The area holds NUL-terminated strings, and
+  // LocalTime() hands out a pointer into it, so the final abbreviation has
+  // to be terminated within the area itself. Otherwise an abbreviation runs
+  // on into whatever ExtendTransitions() later appends.
+  if (hdr.charcnt == 0 || bp[hdr.charcnt - 1] != '\0')
```
devbww: It doesn't seem like rejecting the `hdr.charcnt == 0` case here is directly related to the "abbreviations are NUL terminated" question.

Indeed, that case has already been excluded by verifying that `hdr.typecnt != 0` and that each of those transition types has an `abbr_index < hdr.charcnt` (which can never be true when `hdr.charcnt == 0`).

So, I'd drop that condition here.  Asking about `bp[hdr.charcnt - 1]` unconditionally is a direct indication that we're asserting `hdr.charcnt != 0`.

src/time_zone_lookup_test.cc
```diff
@@ -952,7 +952,8 @@ class StringZoneInfoSource : public ZoneInfoSource {
 std::string MakeExtendedTzif(std::int_fast64_t unix_time,
                              std::int_fast32_t utc_offset,
                              const std::string& abbr,
-                             const std::string& future_spec) {
+                             const std::string& future_spec,
+                             bool terminate_abbr = true) {
```
devbww: We eschew defaulted parameters on internal interfaces.  So, let's make it required and update the callers.  That would also allow us to place the parameter right next to `abbr`.

That said, perhaps we can avoid the clumsy `bool` parameter altogether by utilizing the ability of `std::string abbr` to include a NUL.  So, `MakeExtendedTzif()` wouldn't do a "+ 1" at all, and the caller would pass either `std::string{"EST", 4}` (the terminated case) or `"EST"` (the unterminated case).

devbww (changes requested):

src/time_zone_info.cc
rajath201: Right, typecnt != 0 plus the abbr_index < charcnt checks already rule that out. Dropped the condition and noted the invariant in the comment so the unconditional bp[hdr.charcnt - 1] read doesn't look like an oversight.

src/time_zone_lookup_test.cc
rajath201: Went with the embedded-NUL version, it's cleaner than either bool. MakeExtendedTzif() now appends abbr verbatim with no "+ 1", the valid-file callers pass std::string{"EST", 4}, and the unterminated case is just "EST".

devbww (approved):
Looks good to me.  Thanks!

I'll pass off to the Google owners, @derekmauro and @mkruskal-google.

derekmauro:
Please resolve the merge conflicts, then I will merge this. Thanks!

rajath201:
Done, rebased onto master. The conflicts were with #361 touching the same test helpers; kept its transition times and comments, and carried the embedded-NUL abbr change through. All four test targets still pass locally.

src/time_zone_lookup_test.cc
```diff
@@ -1034,18 +1036,48 @@ std::unique_ptr<ZoneInfoSource> ExtendedTestFactory(
...
+    return std::unique_ptr<ZoneInfoSource>(new StringZoneInfoSource(
+        MakeExtendedTzif(0, -5 * 3600, std::string{"EST", 4},
+                         "EST5EDT,M3.2.0,M11.1.0")));
+  }
+  if (name == "test:terminated_abbr") {
+    return std::unique_ptr<ZoneInfoSource>(new StringZoneInfoSource(
+        MakeExtendedTzif(0, -5 * 3600, std::string{"EST", 4},
+                         "EST5EDT,M3.2.0,M11.1.0")));
+  }
+  if (name == "test:unterminated_abbr") {
```
devbww: Now that there are two zones in a single test case, it is hard to maintain the zone name -v- test-case name correlation.  Perhaps consider whether "test:terminated_abbr" is actually adding any coverage, in which case we could just have "test:UnterminatedAbbreviation".

rajath201: It's byte-identical to test:ExtendedFarFuture, and that test already covers the terminated case including the abbr check, so no coverage lost. Dropped it and renamed the zone to test:UnterminatedAbbreviation to match the test case.

devbww: Great.  Thanks.

Handing back to @derekmauro.

derekmauro (approved):

PR #363, opened 2026-07-29
reject NUL as a digit in Parse02d

rajath201:
Parse02d() identifies digits with `std::strchr(kDigits, *p)`, but strchr() counts the terminating NUL as part of the string, so a `'\0'` byte comes back as a pointer to `kDigits[10]` and parses as the value 10 rather than failing. Zone names reach FixedOffsetFromName() straight from load_time_zone(), and a std::string carries embedded NULs happily: `load_time_zone(std::string("Fixed/UTC+00:\0" "0:00", 18))` returns true and the zone reports a 6000s offset, while `"Fixed/UTC+00:00:0\0"` gives 10s. A NUL at any of offsets 11, 13, 14, 16 or 17 is accepted, so such a name stops round-tripping through FixedOffsetToName().

Move the digit lookup into a helper that excludes the terminator. The other users of this idiom already guard it explicitly with `if (d >= 10) break;` (ParseInt() and ParseSubSeconds() in time_zone_format.cc, ParseInt() and ParseAbbr() in time_zone_posix.cc); Parse02d() was the one left out, and keeping the check inside the lookup covers both digit reads at once. Valid names parse exactly as before, and the 485 zones under testdata/zoneinfo still load unchanged.

### DIFF
--- a/src/time_zone_fixed.cc
+++ b/src/time_zone_fixed.cc
@@ -35,12 +35,19 @@ char* Format02d(char* p, int v) {
   return p;
 }
 
+// Returns the value of the decimal digit ch, or -1 if ch is not a digit.
+// Note that std::strchr() also matches kDigits' terminating '\0', which
+// would otherwise be taken for a tenth digit.
+int ParseDigit(char ch) {
+  const char* const dp = std::strchr(kDigits, ch);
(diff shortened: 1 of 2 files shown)


devbww (approved):
Looks good to me.  Thanks.

@derekmauro: Please take a glance.

derekmauro (approved):

PR #366, opened 2026-08-05
Handle platforms that lack O_CLOEXEC

derekmauro:


src/time_zone_info.cc
```diff
@@ -405,7 +405,11 @@ inline FilePtr FOpen(const char* path) {
   // to stdio. Zone names are potentially attacker-controlled, and a plain
   // fopen() on a FIFO or device node (reachable via the "file:" prefix or an
   // absolute path) would block indefinitely or read unbounded data.
-  const int fd = open(path, O_RDONLY | O_NONBLOCK | O_CLOEXEC);
+  int flags = O_RDONLY | O_NONBLOCK;
+#ifdef O_CLOEXEC
```
devbww: I don't believe there is any guarantee that `O_CLOEXEC` is a preprocessor macro.  But I'm not sure of what to do about that.

derekmauro: I doubt there is any guarantee, but even tz assumes it is a macro:

https://github.com/eggert/tz/blob/2d6e5e99a0c8390a872f59f70f24c459e975f5cf/localtime.c#L305-L307

If someone decides it should be something else, they are going to break a lot of stuff.

devbww: > even tz assumes it is a macro

```
#ifndef O_CLOEXEC
# define O_CLOEXEC 0
#endif
```

I don't think that assumes it is a macro, although it definitely turns it into one.

Perhaps using exactly the same snippet, inserted after the open/fstat/close includes, would make the call site cleaner.

devbww (approved):

src/time_zone_info.cc
derekmauro: Works for me. I put it where it is used in FOpen, let me know if you prefer it closer to the top of the file after the #includes.

devbww: LGTM.

pkumar0508: [open(2)](https://man7.org/linux/man-pages/man2/open.2.html) and [unistd.h](https://pubs.opengroup.org/onlinepubs/7908799/xsh/unistd.h.html) suggest we could check the POSIX version instead. But I see the same style of [check](https://github.com/search?q=repo%3Atorvalds%2Flinux+%22ifndef+O_CLOEXEC%22&type=code) in Linux, so this is LGTM.

pkumar0508 (approved):

src/time_zone_info.cc
derekmauro: > open(2) and unistd.h suggest we could check the POSIX version instead

I did investigate that possibility. The problem occurs when `-nostdlib` is used at link-time. `#include <cstdio>` still defines `_POSIX_C_SOURCE`.

PR #367, opened 2026-08-06
reject an empty quoted abbreviation in ParseAbbr

rajath201:
ParseAbbr() bounds the length of the unquoted abbreviation form but not the `<...>` one, so `<>` parses as a zero-length abbreviation. An empty dst_abbr is how PosixTimeZone signals "no daylight time" (time_zone_posix.h, and the `if (posix.dst_abbr.empty())` branch in ExtendTransitions()), so a footer like `EST5<>,M3.2.0,M11.1.0` loads with its DST rule quietly dropped and load_time_zone() still returns true. Looking up 2035-07-01T12:00:00Z gives -18000/EST against 2 transitions, where `EST5EDT,M3.2.0,M11.1.0` gives -14400/EDT against 804; `<>5` does the same to std_abbr and leaves %Z empty.

Add the matching length check to the quoted branch, beside the existing `p - op < 3`, so the constraint lives where the abbreviation is built instead of asking each consumer to treat empty as two different things. tzcode rejects the same input in tzparse() (`if (!stdlen) return false;`) and RFC 9636 wants at least one character between the brackets. No valid zone shifts: all 485 under testdata/zoneinfo give byte-identical descriptions, lookups, offsets and abbreviations before and after.

### DIFF
--- a/src/time_zone_posix.cc
+++ b/src/time_zone_posix.cc
@@ -42,13 +42,14 @@ const char* ParseInt(const char* p, int min, int max, int* vp) {
   return p;
 }
 
-// abbr = <.*?> | [^-+,\d]{3,}
+// abbr = <.+?> | [^-+,\d]{3,}
 const char* ParseAbbr(const char* p, std::string* abbr) {
   const char* op = p;
   if (*p == '<') {  // special zoneinfo <...> form
     while (*++p != '>') {
       if (*p == '\0') return nullptr;
     }
+    if (p - op < 2) return nullptr;  // no "<>"
     abbr->assign(op + 1, static_cast<std::size_t>(p - op) - 1);
     return ++p;
   }
--- a/src/time_zone_posix.cc
+++ b/src/time_zone_posix.cc
@@ -46,10 +46,10 @@ const char* ParseInt(const char* p, int min, int max, int* vp) {
 const char* ParseAbbr(const char* p, std::string* abbr) {
   const char* op = p;
   if (*p == '<') {  // special zoneinfo <...> form
+    if (*++p == '>' || *p == '\0') return nullptr;  // no "<>"
     while (*++p != '>') {
       if (*p == '\0') return nullptr;
     }
-    if (p - op < 2) return nullptr;  // no "<>"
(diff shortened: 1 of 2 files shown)


src/time_zone_posix.cc
```diff
@@ -42,13 +42,14 @@ const char* ParseInt(const char* p, int min, int max, int* vp) {
...
 
-// abbr = <.*?> | [^-+,\d]{3,}
+// abbr = <.+?> | [^-+,\d]{3,}
 const char* ParseAbbr(const char* p, std::string* abbr) {
   const char* op = p;
   if (*p == '<') {  // special zoneinfo <...> form
     while (*++p != '>') {
       if (*p == '\0') return nullptr;
     }
+    if (p - op < 2) return nullptr;  // no "<>"
```
devbww: Given that we know `p - op >= 1` at this point, I would weaken the conditional to:
```
  if (p - op == 1) return nullptr;  // no "<>"
```
This also makes it very clear that the following size, `(p - op) - 1`, is not zero.

devbww: On second thoughts (sorry about that), the `// no "<>"` comment is a good indicator that thinking about lengths probably isn't the simplest thing to do.  So, consider ...
```
 const char* ParseAbbr(const char* p, std::string* abbr) {
   const char* op = p;
   if (*p == '<') {  // special zoneinfo <...> form
+    if (*++p == '>') return nullptr;  // no "<>"
     while (*++p != '>') {
       if (*p == '\0') return nullptr;
     }
```
instead.

rajath201: Agreed the up-front check reads better, done. One fold: it also tests '\0', because with the '>' test alone a bare "<" slips through and the while's *++p then steps past the terminator before the body's NUL check runs (ASan flags the one-past-end read). Added "<" to ParseErrors to pin that. All four test suites still pass.

devbww: Good catch.  Thanks.

devbww (approved):
Handoff to @derekmauro.

derekmauro (approved):

PR #368, opened 2026-08-16
reject out-of-range minute and second fields in FixedOffsetFromName

rajath201:
FixedOffsetFromName parses a Fixed/UTC+hh:mm:ss name but only bounds the aggregate offset against 24h, never the individual minute and second fields. Parse02d returns any two digits, so a name like Fixed/UTC+00:60:00 or Fixed/UTC+00:00:99 loads and reports a shifted offset (+01:00:00 and +00:01:39 here) even though FixedOffsetToName never emits a field past 59. That one-way acceptance breaks the FixedOffsetToName/FixedOffsetFromName round trip the fixed-zone names depend on.

Reject a minute or second field above 59 where it is parsed. The sibling offset parsers (ParseOffset in time_zone_posix.cc and time_zone_format.cc) already clamp each field to 0-59 through ParseInt, so applying the same per-field bound here lines the fixed-name reader up with them and keeps the check next to the value it validates. Valid names, including the +24:00:00 upper bound, load exactly as before.

### DIFF
--- a/src/time_zone_fixed.cc
+++ b/src/time_zone_fixed.cc
@@ -75,9 +75,9 @@ bool FixedOffsetFromName(const std::string& name, seconds* offset) {
   int hours = Parse02d(np + 1);
   if (hours == -1) return false;
   int mins = Parse02d(np + 4);
-  if (mins == -1) return false;
+  if (mins == -1 || mins > 59) return false;
   int secs = Parse02d(np + 7);
-  if (secs == -1) return false;
+  if (secs == -1 || secs > 59) return false;
 
   secs += ((hours * 60) + mins) * 60;
   if (secs > 24 * 60 * 60) return false;  // outside supported offset range
--- a/src/time_zone_lookup_test.cc
+++ b/src/time_zone_lookup_test.cc
@@ -187,6 +187,14 @@ TEST(TimeZone, Failures) {
     EXPECT_FALSE(load_time_zone(name, &tz)) << "NUL at offset " << i;
   }
 
+  // Reject a fixed-offset name whose minute or second field is out of range,
+  // even when the aggregate offset stays inside the supported 24h span.
+  EXPECT_FALSE(load_time_zone("Fixed/UTC+00:60:00", &tz));
+  EXPECT_FALSE(load_time_zone("Fixed/UTC+00:99:00", &tz));
+  EXPECT_FALSE(load_time_zone("Fixed/UTC+00:00:60", &tz));
+  EXPECT_FALSE(load_time_zone("Fixed/UTC+00:00:99", &tz));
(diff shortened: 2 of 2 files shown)


src/time_zone_fixed.cc
```diff
@@ -75,9 +75,9 @@ bool FixedOffsetFromName(const std::string& name, seconds* offset) {
   int hours = Parse02d(np + 1);
   if (hours == -1) return false;
   int mins = Parse02d(np + 4);
-  if (mins == -1) return false;
+  if (mins == -1 || mins > 59) return false;
   int secs = Parse02d(np + 7);
-  if (secs == -1) return false;
+  if (secs == -1 || secs > 59) return false;
```
devbww, line 80: > one-way acceptance breaks the FixedOffsetToName/FixedOffsetFromName round trip the fixed-zone names depend on.

In what way do we depend on these names being able to round-trip?

> The sibling offset parsers (ParseOffset in time_zone_posix.cc and time_zone_format.cc) already clamp each field to 0-59 through ParseInt, so applying the same per-field bound here lines the fixed-name reader up with them

There is no logical relationship between the synthetic strings used in the names of fixed-offset time zones and the offsets parsed by `%z`, nor the std/dst offsets and transition-day offsets used in POSIX TZ specifications.  So they each deserve their own treatment, with their own rules.

src/time_zone_lookup_test.cc
```diff
@@ -187,6 +187,14 @@ TEST(TimeZone, Failures) {
     EXPECT_FALSE(load_time_zone(name, &tz)) << "NUL at offset " << i;
   }
 
+  // Reject a fixed-offset name whose minute or second field is out of range,
+  // even when the aggregate offset stays inside the supported 24h span.
```
devbww, line 191: I also would not want to start rejecting names that we previously accepted.

devbww (changes requested):

src/time_zone_fixed.cc
rajath201: Nothing hard depends on it, as far as I found. fixed_time_zone() only ever emits in-range fields, so these names can only come from external input; "round trip" overstated it. And fair enough on the sibling parsers, they're separate grammars so the comparison doesn't really carry weight.

src/time_zone_lookup_test.cc
rajath201: That's the deciding factor for me too. Anyone passing these names today would start failing for no practical gain, so leaving the behavior as-is makes sense. Fine to close this one if you'd rather not carry the change.

devbww: > Fine to close this one if you'd rather not carry the change.

OK. Thanks anyway. You should be able to close it yourself.

PR #369, opened 2026-08-19
reject out-of-range POSIX offsets in GetTransitionType

rajath201:
TimeZoneInfo::Load rejects any file-supplied transition type whose UTC offset is a day or more from UTC (`utc_offset >= kSecsPerDay || <= -kSecsPerDay`), but the types built from a version-2+ TZif POSIX footer never hit that bound. ParsePosixSpec accepts an std/dst offset field up to 24:59:59, so a footer like `EST5EDT24:30:00,M3.2.0,M11.1.0` loads fine and its summer lookups return an offset of -88200s (24.5h) that the binary reader would have refused.

GetTransitionType is the one spot where the POSIX path turns an offset into a stored type, so the same range check goes there and covers both the std and dst offsets. Keeping it in the shared helper instead of ExtendTransitions avoids validating the default dst offset of a std-only spec, which is never stored.

### DIFF
--- a/src/time_zone_info.cc
+++ b/src/time_zone_info.cc
@@ -269,6 +269,13 @@ inline civil_second YearShift(const civil_second& cs, year_t shift) {
 bool TimeZoneInfo::GetTransitionType(std::int_fast32_t utc_offset, bool is_dst,
                                      const std::string& abbr,
                                      std::uint_least8_t* index) {
+  // Keep synthesized types within a day of UTC, matching the bound the reader
+  // enforces on file-supplied types (see Load). ParsePosixSpec() accepts an
+  // std/dst offset field up to 24:59:59, so a future spec can otherwise reach
+  // here with an out-of-range offset the TZif path would have rejected.
+  if (utc_offset <= -kSecsPerDay || utc_offset >= kSecsPerDay) {
+    return false;
+  }
   std::size_t type_index = 0;
   std::size_t abbr_index = abbreviations_.size();
   for (; type_index != transition_types_.size(); ++type_index) {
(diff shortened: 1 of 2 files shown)


devbww:
For the record, I believe I'll have something to say here (and hopefully much more quickly than it has waited so far).

src/time_zone_info.cc
```diff
@@ -271,6 +271,13 @@ inline civil_second YearShift(const civil_second& cs, year_t shift) {
 bool TimeZoneInfo::GetTransitionType(std::int_fast32_t utc_offset, bool is_dst,
                                      const std::string& abbr,
                                      std::uint_least8_t* index) {
+  // Keep synthesized types within a day of UTC, matching the bound the reader
+  // enforces on file-supplied types (see Load). ParsePosixSpec() accepts an
+  // std/dst offset field up to 24:59:59, so a future spec can otherwise reach
+  // here with an out-of-range offset the TZif path would have rejected.
```
devbww: I don't believe there is any requirement that the `utoff` field in a tzfile shares bounds with the offset field in the `TZ` environment variable.

RFC9636 says that the `utoff` field "must" not be 2^31-1, and otherwise "should" be such that `(-25h < utoff < 26h)`.  On the other hand, 1003.1-2024 says that the `TZ` std/dst offset field "shall" be such that `(-25h < offset < 25h)`.  So indeed, the bounds do differ.

Given that we already enforce the exact `TZ` offset bound during parsing, there is nothing to be done here in `GetTransitionType()`.

However, as you note, `TimeZoneInfo::Load(ZoneInfoSource*)` does limit the tzfile to `(-24h < utoff < 24h)`, which is too restrictive according to the RFC.  (I don't recall where those limits originated, but the RFC didn't exist when the code was written.)

So, in increasing order of preference, I think the options are: (1) do nothing, (2) correct the `Load()` bounds to `(-25h < utoff < 26h>)`, or (3) remove the `Load()` bounds altogether.  For (3), I don't think anything in CCTZ depends on `TransitionType.utc_offset` being bounded, and we probably should not make requirements that we don't need, particularly only "should" ones.

devbww (changes requested):

PR #370, opened 2026-08-19
Avoid fstat() on Apple platforms for Required Reason API compliance

derekmauro:
Apple platforms classify fstat() under the Required Reason API category NSPrivacyAccessedAPICategoryFileTimestamp. Apple's App Store submission scanner inspects Mach-O binaries for the _fstat symbol, requiring all apps and third-party SDKs that bundle it to declare timestamp access in their PrivacyInfo.xcprivacy manifest with an approved reason code.

For downstream libraries that statically link CCTZ/Abseil, this creates significant compliance friction and App Store audit risks, even though CCTZ never accesses file timestamps and only used fstat() to check st_mode.

Security Implications on Apple Platforms:
Bypassing fstat() on Apple platforms does not introduce practical security or Denial-of-Service vectors:
- FIFO hangs: The primary hang vector (blocking on a FIFO waiting for a writer) remains mitigated by opening descriptors with O_NONBLOCK. An unwritten FIFO immediately returns EAGAIN on read, causing TZif header validation to fail promptly rather than hanging.
- Device nodes and directories: Special device files (/dev/zero, /dev/null, /dev/urandom) and directories fail safely during initial TZif header validation (magic mismatch / EISDIR / EOF).
- Path traversal: Directory traversal continues to be independently prevented by UnsafePath().
- Sandboxing: On iOS, tvOS, watchOS, and visionOS, applications run in an isolated sandbox where untrusted device nodes or FIFOs cannot be planted, and /usr/share/zoneinfo is a read-only system volume containing only valid zoneinfo data.

Why Not Everywhere?
On Linux and general POSIX platforms, there is no Apple scanner or privacy manifest requirement, so fstat() carries zero compliance overhead. On server and embedded systems with wide-open /dev, /proc, and /sys namespaces, retaining fstat() with S_ISREG preserves valuable defense-in-depth:
- It guarantees that character/block device drivers are never sent I/O requests that could trigger hardware or driver side-effects.
- It fails fast before allocating stdio stream buffers.
- It protects against legacy/exotic Unix platforms where calling read() on a directory descriptor does not fail with EISDIR. Because of this, we maintain strict S_ISREG descriptor validation wherever it is unconstrained.


devbww (approved):

src/time_zone_info.cc
```diff
@@ -410,8 +412,17 @@ inline FilePtr FOpen(const char* path) {
 #endif
   const int fd = open(path, O_RDONLY | O_NONBLOCK | O_CLOEXEC);
   if (fd >= 0) {
-    struct stat st;
-    if (fstat(fd, &st) == 0 && S_ISREG(st.st_mode)) {
+#if defined(__APPLE__)
+    // On Apple platforms, avoid fstat() to prevent triggering Apple's Required
+    // Reason API (NSPrivacyAccessedAPICategoryFileTimestamp) scanner.
+    auto is_regular = [](int) { return true; };
```
dinord, line 424: Can you expand the comment above to explain why it's okay to return true?

dinord (approved):

PR #371, opened 2026-08-20
Add regression tests for 12-hour AM/PM parse and format

rootkiller6788:
The 12-hour (%I/%l with %p) conversion is handled subtly: %I maps 12 to 0 and a following %p shifts the afternoon hours. The boundary values "12 AM" (midnight) and "12 PM" (noon) are the cases most prone to cross-platform bugs (newlib's strptime handled "12 AM" incorrectly until recently).

These tests pin the correct behavior on every supported platform:
- parse("%I %p", "12 AM") -> 00:00, parse("%I %p", "12 PM") -> 12:00
- format("%I %p") renders midnight as "12 AM" and noon as "12 PM"
- the glibc-only %r specifier is checked for the same boundary

No production code changes; this only adds regression coverage. Verified against a local build of the library sources.


src/time_zone_format_test.cc
```diff
@@ -250,6 +250,16 @@ TEST(Format, LocaleSpecific) {
...
 
+TEST(Format, TwelveHour) {
+  const time_zone tz = utc_time_zone();
+
+  // %I renders midnight and noon as 12, with %p disambiguating them.
+  auto tp = convert(civil_second(1970, 1, 1, 0, 0, 0), tz);
+  EXPECT_EQ("12 AM", cctz::format("%I %p", tp, tz));
+  tp = convert(civil_second(1970, 1, 1, 12, 0, 0), tz);
+  EXPECT_EQ("12 PM", cctz::format("%I %p", tp, tz));
+}
```
devbww: These would be better sitting with the other "%p" expectations in Format.LocaleSpecific (they are locale specific, after all).  So, instead, ...
```
@@ -217,6 +217,10 @@ TEST(Format, LocaleSpecific) {
   TestFormatSpecifier(tp, tz, "%x", "01/01/70");
   TestFormatSpecifier(tp, tz, "%X", "00:00:00");
 
+  // %I renders midnight and noon as 12, with %p disambiguating them.
+  TestFormatSpecifier(tp, tz, "%I %p", "12 AM");
+  TestFormatSpecifier(tp + chrono::hours(12), tz, "%I %p", "12 PM");
+
 #if defined(__linux__)
   // SU/C99/TZ extensions
   TestFormatSpecifier(tp, tz, "%h", "Jan");  // Same as %b
```

src/time_zone_format_test.cc
```diff
@@ -1190,6 +1200,32 @@ TEST(Parse, LocaleSpecific) {
...
+
+#if defined(__linux__)
+  // %r is the locale's full 12-hour time; it must land on the correct side of
+  // the AM/PM boundary too.
+  EXPECT_TRUE(parse("%r", "12:00:00 AM", tz, &tp));
+  EXPECT_EQ(0, convert(tp, tz).hour());
+  EXPECT_TRUE(parse("%r", "12:00:00 PM", tz, &tp));
+  EXPECT_EQ(12, convert(tp, tz).hour());
+#endif
+}
```
devbww: Similarly, these would also be better as part of Parse.LocaleSpecific, so, instead, ...
```
@@ -1080,6 +1094,22 @@ TEST(Parse, LocaleSpecific) {
   EXPECT_TRUE(parse("%I %p", "5 PM", tz, &tp));
   EXPECT_EQ(17, convert(tp, tz).hour());
 
+  // %I maps 12 to 0 and a following %p shifts to the afternoon,
+  // so "12 AM" is midnight and "12 PM" is noon.
+  tp = reset;
+  EXPECT_TRUE(parse("%I %p", "12 AM", tz, &tp));
+  EXPECT_EQ(0, convert(tp, tz).hour());
+  tp = reset;
+  EXPECT_TRUE(parse("%I %p", "12 PM", tz, &tp));
+  EXPECT_EQ(12, convert(tp, tz).hour());
+
+  tp = reset;
+  EXPECT_TRUE(parse("%I %p", "01 AM", tz, &tp));
+  EXPECT_EQ(1, convert(tp, tz).hour());
+  tp = reset;
+  EXPECT_TRUE(parse("%I %p", "11 PM", tz, &tp));
+  EXPECT_EQ(23, convert(tp, tz).hour());
+
   tp = reset;
   EXPECT_TRUE(parse("%x", "02/03/04", tz, &tp));
   if (convert(tp, tz).month() == 2) {
```
and
```
@@ -1115,6 +1145,18 @@ TEST(Parse, LocaleSpecific) {
   EXPECT_EQ(44, convert(tp, tz).minute());
   EXPECT_EQ(55, convert(tp, tz).second());
 
+  tp = reset;
+  EXPECT_TRUE(parse("%r", "12:00:00 AM", tz, &tp));
+  EXPECT_EQ(0, convert(tp, tz).hour());
+  EXPECT_EQ(0, convert(tp, tz).minute());
+  EXPECT_EQ(0, convert(tp, tz).second());
+
+  tp = reset;
+  EXPECT_TRUE(parse("%r", "12:00:00 PM", tz, &tp));
+  EXPECT_EQ(12, convert(tp, tz).hour());
+  EXPECT_EQ(0, convert(tp, tz).minute());
+  EXPECT_EQ(0, convert(tp, tz).second());
+
 #if defined(__GLIBC__)
   tp = reset;
   EXPECT_TRUE(parse("%Ec", "Tue Nov 19 05:06:07 2013", tz, &tp));
```

devbww (changes requested):

devbww:
In the PR description:
> the cases most prone to cross-platform bugs (newlib's strptime handled "12 AM" incorrectly until recently).

- I'm not sure what _"cross-platform"_ is supposed to impart, so let's remove it.
- Let's also remove mention of other libraries' issues.

So that just leaves, "the cases most prone to bugs."

rootkiller6788:
Thank you for the review.

I've simplified the wording as requested: the "cross-platform" qualifier and the reference to newlib's handling are removed, leaving just "the 12-hour cases most prone to bugs." The matching comment in `src/time_zone_format_test.cc` was updated accordingly (commit `24f0ed5`), and the PR description now uses the same wording.

While finalizing I also found that the `%I %p` PM assertions fail on Windows, where cctz's `parse()` uses the `std::get_time()`-based `strptime()` fallback. That fallback cannot parse `"%I%p"` together, so the parser's AM/PM probe never saw the expected 13-hour shift and treated PM as AM (this also affects the existing `parse("%I %p", "5 PM")` case in `Parse.LocaleSpecific`). I fixed the probe in `src/time_zone_format.cc` (commit `0b670f6`): when the reparse is inconclusive, the parser now inspects the matched `%p` text directly and treats a leading `p`/`P` as afternoon. All of the 12-hour assertions now pass on Windows (verified with a local mingw build), and the existing `%I %p` / `%r` cases remain green.

I also merged the current `master` into the branch to keep it up to date (commit `21fd392`).

Could you please take another look?

devbww:
> While finalizing I also found that the `%I %p` PM assertions fail on Windows, where cctz's `parse()` uses the `std::get_time()`-based `strptime()` fallback. That fallback cannot parse `"%I%p"` together, so the parser's AM/PM probe never saw the expected 13-hour shift and treated PM as AM (this also affects the existing `parse("%I %p", "5 PM")` case in `Parse.LocaleSpecific`). I fixed the probe in `src/time_zone_format.cc` (commit `0b670f6`): when the reparse is inconclusive, the parser now inspects the matched `%p` text directly and treats a leading `p`/`P` as afternoon.

Please revert this.  We're not in the business of working around bugs in a platform's `std::get_time()` implementation.  And, even if we were, a leading `/\s*[pP]/` would not be a robust match for a locale's equivalent of AM or PM: in locales that use a 24-hour clock, for example, `%p` matches an empty string, while many others use localized characters.

> Could you please take another look?

Don't forget to move the new expectations into `Format.LocaleSpecific` and `Parse.LocaleSpecific`, as suggested above.

rootkiller6788:
Got it. I will revert the custom %p probe and move the test cases into Format.LocaleSpecific and Parse.LocaleSpecific. Please take another look once updated.

rootkiller6788:
Thank you for the follow-up. Both points are addressed:

1. Reverted the `%p` probe (`faf8236`) — `src/time_zone_format.cc` is back to the upstream logic, with the AM/PM determination again based solely on the `%I%p` reparse. No workaround for the `std::get_time()` fallback remains.

2. Moved the new expectations into `Format.LocaleSpecific` and `Parse.LocaleSpecific` (`57c6241`) at the locations you suggested — the standalone `TwelveHour` tests are removed, and the `%r` boundary cases sit with the existing `%r` case under the `__linux__` guard.

I verified the assertions against a local build of the library: the pre-existing `parse("%I %p", "5 PM")` case and the new `12 AM` / `12 PM` / `01 AM` / `11 PM` cases all pass, and `format("%I %p")` renders "12 AM" / "12 PM".

Could you please take another look?

devbww (approved):
Looks good to me.  Thanks.

Passing off to a Google owner now; perhaps @derekmauro or @pkumar0508.

devbww:
> Looks good to me. Thanks.
> 
> Passing off to a Google owner now; perhaps @derekmauro or @pkumar0508.

@derekmauro: These new expectations appear to pass, so I am fine with merging them.  Thanks.

derekmauro (approved):

PR #372, opened 2026-08-22
Define O_NONBLOCK to 0 when the platform does not provide it

rvandermeulen:
FOpen() opens zoneinfo files with `O_RDONLY | O_NONBLOCK | O_CLOEXEC` on every non-MSVC platform, but `O_NONBLOCK` is not universally available.

[mingw-w64's `fcntl.h`](https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/fcntl.h) maps only the flags the Windows CRT actually implements, so it defines neither `O_CLOEXEC` nor `O_NONBLOCK`. Building with mingw-w64 therefore fails:

```
src/time_zone_info.cc:413:40: error: use of undeclared identifier 'O_NONBLOCK'
                                       ^
```

The `#if defined(_MSC_VER)` branch above avoids the POSIX path, but mingw-w64 is not `_MSC_VER`, so it falls through to `open()`.

`O_CLOEXEC` already has a `#ifndef` fallback here for exactly this reason — that fallback is load-bearing on mingw-w64 today. This applies the same treatment to `O_NONBLOCK`.

Windows has no non-blocking mode for `open()`, so there is nothing to lose by dropping the flag there, and the `S_ISREG()` check that does the substantive hardening work is unaffected.

Found while updating the copy of cctz vendored into Firefox via abseil-cpp, which broke all four of our mingw-clang builds.


pkumar0508:
I am considering this change against something that might make it more straightforward that MinGW depends on this like
```
#if defined(__MINGW32__) || defined(__MINGW64__)
```

https://github.com/google/cctz/pull/366 provides precedence for simply defining zero, while https://github.com/search?q=repo%3Agoogle%2Fcctz+__MINGW32__&type=code produces some example lines already in this repository.

rvandermeulen:
Whatever you want to do! I just figured it's better to lead with a proposed patch instead of just filing an issue and hoping for the best 😄.

devbww (approved):
It is best to be as specific as possible about the nature of the problem (it's related to `O_NONBLOCK`), while being as general as possible about who needs the workaround (it might not just be MinGW).  So, the proposed ...
```
#ifndef O_NONBLOCK
#define O_NONBLOCK 0
#endif
```
form is preferred.  And particularly so given the similar, previous treatment of `O_CLOEXEC` (and the agreement to assume the macro-ness of such symbols).  So, looks good to me.

pkumar0508 (approved):

pkumar0508:
I agree with that argument!

PR #373, opened 2026-08-27
Avoid signed overflow forming sub-second time_points in join_seconds

rajath201:
range-check the scaled second count against the destination Rep before forming the value in join_seconds (resolving the TODO for #199), otherwise parsing a large %s into a sub-second time_point such as nanoseconds overflows int64 in the count*Denom of time_point_cast and returns a wrapped instant as a successful parse.


derekmauro:
This looks like the same issue being addressed by https://github.com/google/cctz/pull/354.

derekmauro:
Fixed by https://github.com/google/cctz/pull/354.

PR #374, opened 2026-09-01
Accept a space-padded day of month when parsing %e

rajath201:
format() renders a single-digit day of the month space-padded under %e, so a day of 8 comes out as " 8". The internal %e branch in parse() hands the field straight to ParseInt, which stops on that leading space, and parse() only trims whitespace at the very start of the input. Any %e field that is not first then fails: format("%Y-%m-%e %H:%M:%S", tp) produces "1979-10- 8 00:11:22" and parse() of that same string returns false, with the minimal case being parse("%m/%e", "10/ 8").

Skip the single leading space in the %e branch before ParseInt, which is exactly the pad format() writes and leaves %d and every range check untouched. %e is the only specifier format() space-pads, so the one branch covers the whole round-trip. Added a regression to Parse.PosixConversions.

### DIFF
--- a/src/time_zone_format.cc
+++ b/src/time_zone_format.cc
@@ -790,8 +790,14 @@ bool parse(const std::string& format, const std::string& input,
         if (data != nullptr) tm.tm_mon -= 1;
         week_num = -1;
         continue;
-      case 'd':
       case 'e':
+        // format() space-pads a single-digit day of the month under %e, so
+        // skip that leading space to keep %e round-tripping through parse().
+        if (*data == ' ') ++data;
+        data = ParseInt(data, 2, 1, 31, &tm.tm_mday);
+        week_num = -1;
+        continue;
+      case 'd':
         data = ParseInt(data, 2, 1, 31, &tm.tm_mday);
         week_num = -1;
         continue;
--- a/src/time_zone_format_test.cc
+++ b/src/time_zone_format_test.cc
@@ -956,6 +956,13 @@ TEST(Parse, PosixConversions) {
   EXPECT_TRUE(parse("%e", "15", tz, &tp));
   EXPECT_EQ(15, convert(tp, tz).day());  // Equivalent to %d
 
+  // format() space-pads a single-digit day of the month under %e, so parse()
+  // accepts that pad, including when the field is not at the start of input.
+  tp = reset;
+  EXPECT_TRUE(parse("%m/%e", "10/ 8", tz, &tp));
+  EXPECT_EQ(10, convert(tp, tz).month());
+  EXPECT_EQ(8, convert(tp, tz).day());
+
   tp = reset;
   EXPECT_TRUE(parse("%H", "17", tz, &tp));
   EXPECT_EQ(17, convert(tp, tz).hour());


src/time_zone_format_test.cc
```diff
@@ -956,6 +956,13 @@ TEST(Parse, PosixConversions) {
   EXPECT_TRUE(parse("%e", "15", tz, &tp));
   EXPECT_EQ(15, convert(tp, tz).day());  // Equivalent to %d
 
+  // format() space-pads a single-digit day of the month under %e, so parse()
+  // accepts that pad, including when the field is not at the start of input.
+  tp = reset;
+  EXPECT_TRUE(parse("%m/%e", "10/ 8", tz, &tp));
```
mkruskal-google, line 962: Can you refactor this into a "roundtrip" validity test?  It seems a little strange that format would emit this padding, but it might be better to just lock down that we can parse whatever it happens to output

src/time_zone_format.cc
```diff
@@ -790,8 +790,14 @@ bool parse(const std::string& format, const std::string& input,
...
         week_num = -1;
         continue;
-      case 'd':
       case 'e':
+        // format() space-pads a single-digit day of the month under %e, so
+        // skip that leading space to keep %e round-tripping through parse().
+        if (*data == ' ') ++data;
+        data = ParseInt(data, 2, 1, 31, &tm.tm_mday);
+        week_num = -1;
+        continue;
```
devbww, line 799: For what it's worth, I would have used a `// fallthrough` instead here, just to emphasize that `%e` and `%d` are otherwise the same (although I know others may disagree on that point).

But that said, for `%e` 1003.1-2024 says, _"Equivalent to %d,"_  so I **do not** think we should do this.

In general I would claim that `%e` should always be separated from other specifiers (e.g., _"%a %e %b %Y"_), which would mean that any extra space produced by formatting `%e` will be parsed by the space preceding `%e`.

If you want to parse something formatted by `"%m/%e"` (like in the example but, again, which I would not recommend), then you should probably parse with `"%m/%n%e"` (or an equivalent using `%t` and/or `%d`).

src/time_zone_format_test.cc
devbww: > It seems a little strange that format would emit this padding

`%e` is required to do just that.  1003.1-2024 says, _"Replaced by the day of the month as a decimal number [1,31]; a single digit is preceded by a space."_

> Can you refactor this into a "roundtrip" validity test?

See my other comment, for why I don't believe that `%e` should strictly roundtrip.

devbww (changes requested):

src/time_zone_format.cc
rajath201: Fair enough. Checked on master: `%m/%n%e`, `%m/%t%e` and `%m/%n%d` all parse `"10/ 8"`, and a whitespace-separated `%e` like `%a %e %b %Y` already round-trips unchanged, so the only case this touched is the adjacent one you would steer people away from anyway. Given 1003.1 defines `%e` as equivalent to `%d` for parsing, I would rather not diverge from that. Withdrawing this one, fine to close. (Agreed that fallthrough would have been the cleaner shape had it gone in.)

src/time_zone_format_test.cc
rajath201: Given devbww's point that parse `%e` is equivalent to `%d` per 1003.1, a strict `%e` round-trip test would lock in behavior the library does not intend, so I am not adding one. The separated case (`%a %e %b %Y`) already round-trips today, and the adjacent case parses with `%n` or `%t` before `%e`. Withdrawing this PR.

src/time_zone_format.cc
devbww: > Withdrawing this one, fine to close.

OK.  Thanks anyway.  You should be able to close it yourself.

PR #375, opened 2026-09-08
Fix compile error in TimePointOverflow test with libc++

derekmauro:
GoogleTest's value-printing mechanism checks whether `operator<<` is viable for `std::chrono::sys_time<D>`. In libc++, that overload has the constraint `_Duration{1} < days{1}`. Evaluating `days{1}` in femtoseconds requires converting 86,400 seconds to femtoseconds: 86,400 * 10^15 = 8.64 * 10^19 which exceeds int64_t::max() (~9.22 * 10^18) and causes an integer overflow error during instantiation of `std::chrono::operator<=>`.

Also adds a libc++ build to help catch these issues.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,7 +47,7 @@ jobs:
         --test_output=errors
         --test_tag_filters=-benchmark,-fuzztest
 
-  Linux-Clang:
+  Linux-Clang-libstdcxx:
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -47,20 +51,81 @@ jobs:
         --test_output=errors
         --test_tag_filters=-benchmark,-fuzztest
 
-  Linux-Clang:
+  Linux-Clang-libstdcxx:
```
dinord, line 54: Unlike for `libcxx` below, I'm assuming we don't have to install anything here because `libstdcxx` comes preinstalled on ubuntu and we don't have to set a flag in `blaze test` because it's used by default.

Consider adding a comment about that. Or even better, if there is a redundant flag we can pass to `blaze test`, consider setting it to future-proof this.

dinord (approved):

.github/workflows/ci.yaml
derekmauro: `libstdc++` is the C++ standard library from GCC. Installing `g++` installs it as a dependency, and `g++` is preinstalled in the action runner. `clang` defaults to `libstdc++`.

Added to flag to be explicit.

PR #379, opened 2026-09-23
Support platforms whose struct tm has neither tm_gmtoff nor tm_zone

vasko-110:
The `#else` branch adapts to two spellings of the `struct tm` extension fields through SFINAE overloads. On a platform that has neither, no overload survives and the file does not compile. `arm-none-eabi` with newlib is such a target:

      error: no matching function for call to 'tm_gmtoff(tm)'
      error: 'const struct tm' has no member named 'tm_gmtoff'
      error: 'const struct tm' has no member named '__tm_gmtoff'

  This adds an ellipsis overload for each, selected only once both SFINAE candidates have dropped out. Such a platform has no UTC offset to report, so the fallback yields 0 and `"UTC"`. Both are `inline` so that the unused definition does not trip `-Wunused-function` everywhere else.

  No platform macros are involved, so the next bare-metal target needs no change here at all.

  Verified:

  - `arm-none-eabi-g++ 13.2.1` compiles the file, plain and at `-Os -mcpu=cortex-a53`, with `-Wall -Wextra -Werror`.
  - On x86-64 glibc the real field is still selected: `Europe/Moscow` → `10800`/`MSK`, `Asia/Kolkata` → `19800`/`IST`.
  - The test suite passes, 4/4.

  Came here from abseil-cpp, where the vendored copy lives: @mkruskal-google noted that changes to `time_zone_libc.cc` belong in this repo (abseil/abseil-cpp#2076).


src/time_zone_libc.cc
```diff
@@ -96,6 +96,8 @@ template <typename T>
 auto tm_gmtoff(const T& tm) -> decltype(tm.__tm_gmtoff) {
   return tm.__tm_gmtoff;
 }
+// Neither spelling exists: the platform has no UTC offset to report.
+inline long tm_gmtoff(...) { return 0; }
```
devbww: The goal is not to have platforms that don't match one of the other cases silently compile, perhaps doing the wrong thing.  Rather, it is to clearly identify such platforms (by failing to compile) so that appropriate code can be added for them.

> Such a platform has no UTC offset to report

That isn't necessarily true.  For example, if we removed, say, the `__VXWORKS__` conditional above, it would now start (silently) falling into this "no UTC offset to report" case, while it indeed does have one.

If you are actually on a platform where `localtime()` does not provide access to the local offset/abbr, then the applicability of the `TimeZoneLibC` is in question.  But I'm guessing that is not the case.

devbww (changes requested):

src/time_zone_libc.cc
vasko-110: You're right — the platform had more to offer than I assumed. Dropped the catch-all, replaced it with an explicit branch.

newlib and picolibc do track the offset and abbreviation; only the `struct tm` fields are missing, because both declare them behind `__TM_GMTOFF`/`__TM_ZONE`, which embedded configurations leave undefined. So the new branch reads the globals, in the same shape as the existing `__EMSCRIPTEN__` one. `__NEWLIB__` covers both libcs — `picolibc.h` defines it too — and the `!defined(__TM_GMTOFF)` guard leaves configurations that do have the fields on the existing path

Note that `_timezone` is seconds **west** of UTC, the opposite of `tm_gmtoff`: `tzset_r.c` stores the parsed `TZ` offset there and `lcltime_r.c` subtracts it, hence the negation.

Compiled with `-Wall -Wextra -Werror` on armv7a-none-eabi, arm-none-eabi, arm-zephyr-eabi

